PR #33, opened 2015-10-05
Create interactive_guess.py

Pitje06:
Created tools folder and interactive_fit.py; which contains (mainly) the classes InteractiveFit2D and InteractiveFit3D. These classes allow you to graphically create an initial guess for symfit, which is particularly valuable for complex and chaotic models.

Currently it is possible to use simple 2D models with InteractiveFit2D and multidimensional models with InteractiveFit3D. InteractiveFit2D should be extended to project the higher dimensions into multiple plots in 2 dimensions the same way InteractiveFit3D does in 3 dimensions.

@tBuLi: this was made some time ago, and works for symfit 0.2.6 on Python 2.7 and 3.4 (should anyway). I'm not sure the interal API changed/is used correctly/has impending changes coming.


tools/interactive_fit.py
```diff
@@ -0,0 +1,359 @@
...
+    """A class that provides a visual_guess method which provides
+    an graphical, interactive way of guessing initial fitting parameters."""
+    def _update_plot(self, val):
+        """Callbak to redraw the plot to reflect the new parameter values."""
+        # Since all sliders call this same callback without saying who they are
+        # I need to update the values for all parameters. This can be
+        # circumvented by creating a seperate callback function for each
+        # parameter.
+        for p in self.params:
+            # p.value = self.sliders[p].val?
```
tBuLi: whats wrong with this line? Seems like the cleaner way to do it.

tools/interactive_fit.py
```diff
@@ -0,0 +1,359 @@
...
+        these are not provided, the minimum is 0; and the maximum is value*2.
+        If no initial value is provided, it defaults to 1."""
+
+        x_min = np.min(self.xdata)
+        x_max = np.max(self.xdata)
+        # It would be nice if I could get these from model someway...
+        y_min = 0
+        y_max = 1
+
+        if len(self.vars) != 1:
```
tBuLi: In symfit 0.3.0, this should be self.model.independent_vars

tools/interactive_fit.py
```diff
@@ -0,0 +1,359 @@
...
+        """Create a matplotlib window with sliders for all parameters
+        in this model, so that you may graphically guess initial fitting
+        parameters. n_points is the number of points drawn for the plot.
+        Data points are plotted as blue points, the proposed model as
+        a red line.
+        Slider extremes are taken from the parameters where possible. If
+        these are not provided, the minimum is 0; and the maximum is value*2.
+        If no initial value is provided, it defaults to 1."""
+
+        x_min = np.min(self.xdata)
```
tBuLi: In symfit 0.3.0 self.independent_data is the one you should use here, and it's a dict

tools/interactive_fit.py
```diff
@@ -0,0 +1,359 @@
...
+        x_max = np.max(self.xdata)
+        # It would be nice if I could get these from model someway...
+        y_min = 0
+        y_max = 1
+
+        if len(self.vars) != 1:
+            raise IndexError("Only supports 2D problems!")
+
+        ax = plt.subplot(111)
+        plt.subplots_adjust(bottom=0.1 + 0.05*len(self.params))
```
tBuLi: self.model.params

tools/interactive_fit.py
```diff
@@ -0,0 +1,359 @@
...
+        plt.title("Fit")
+
+        self.xpoints = np.linspace(x_min, x_max, n_points)
+
+        plt.ylim(y_min, y_max)  # Y-Domain for self.model
+        plt.xlim(x_min, x_max)  # X-domain for self.model
+
+        # self.vec_func should probably be replaced for self.model.
+        self.vec_func = lambdify(self.vars+self.params, self.model, "numpy")
```
tBuLi: Don't use lambdafy directly in case of future API changes use sympy_to_py from symfit.core.support instead. Although in this case as mentioned in your comment, just call self.model with whatever args and kwargs you want.

tools/interactive_fit.py
```diff
@@ -0,0 +1,359 @@
...
+        # It might be better to use keyword arguments (p.name: p.value), but
+        # I don't know (any more) if that's supported by self.model.
+        vals = self.vec_func(self.xpoints, *list(p.value for p in self.params))
```
tBuLi: as a general remark, everything to do with the model has nopw been moved to Model objects. so self.params becomes self.model.params

tools/interactive_fit.py
```diff
@@ -0,0 +1,359 @@
...
+        self.sliders = {}
+        for p in self.params:
+            if not p.fixed:
+                axbg = 'lightgoldenrodyellow'
+            else:
+                axbg = 'red'
+            # start-x, start-y, width, height
+            ax = plt.axes([0.162, i, 0.68, 0.03], axisbg=axbg)
+
+            if not hasattr(p, "value") or p.value is None:
```
tBuLi: Value is 1.0 by default. And although I guess it's possible for some yerk to put p.value to None, I think you can just assume it to be a float or int. And you don't have to check if any of these attributes exist. They always do.

tBuLi:
As a general remark, could you also include an example usage and some tests? I know testing with matplotlib can be a pain, but at least it should be possible to check datatypes at some key points.

Could you add them to tests/tests_interactive.py?

tools/interactive_fit.py
```diff
@@ -0,0 +1,359 @@
...
+        plt.ylim(y_min, y_max)  # Y-Domain for self.model
+        plt.xlim(x_min, x_max)  # X-domain for self.model
+
+        # self.vec_func should probably be replaced for self.model.
+        self.vec_func = lambdify(self.vars+self.params, self.model, "numpy")
+        # It might be better to use keyword arguments (p.name: p.value), but
+        # I don't know (any more) if that's supported by self.model.
+        vals = self.vec_func(self.xpoints, *list(p.value for p in self.params))
+        self.my_plot, = ax.plot(self.xpoints, vals, color='red')
+        ax.scatter(self.xdata, self.ydata, c='blue')
```
tBuLi: self.ydata -> self.dependent_data, which is still a dict

tools/interactive_fit.py
```diff
@@ -0,0 +1,359 @@
...
+
+        plt.show()
+
+
+class ProjectionPlot:
+    """Helper class for InteractiveFit3D. It holds information for a plot of
+    one projection in 3 dimensions.
+    Data is plotted as a scatterplot in blue, the model as a red surface."""
+    def __init__(self, axes, xydata, z_interpolator, xymesh, z_function, title):
+        """axes: The axes upon which to plot
```
tBuLi: :param axes: ;)

And shouldn't a title be optional?

tools/interactive_fit.py
```diff
@@ -0,0 +1,359 @@
...
+
+
+if __name__ == "__main__":
+    def f(xy, a, b):
+        x, y, a = xy
+        return np.cos(a*x) * b*np.sin(y)
+
+    from symfit.api import Parameter, Variable, sin, cos
+
+    def f_sym(x, y, a, b):
```
tBuLi: please start using symfit like your supposed to? :P

tools/interactive_fit.py
```diff
@@ -0,0 +1,359 @@
...
+        self._update_plot(self.params[0])  # Update all plots.
+        plt.show()
+
+
+if __name__ == "__main__":
+    def f(xy, a, b):
+        x, y, a = xy
+        return np.cos(a*x) * b*np.sin(y)
+
+    from symfit.api import Parameter, Variable, sin, cos
```
tBuLi: symfit.api -> symfit

tools/interactive_fit.py
```diff
@@ -0,0 +1,359 @@
...
+    xdata = np.linspace(-np.pi, np.pi, 7)
+    ydata = np.linspace(-np.pi, np.pi, 7)
+    adata = np.linspace(0, 2.5, 7)
+
+    xx, yy, aa = np.meshgrid(xdata, ydata, adata)
+
+    xydata = np.array((xx.flatten(), yy.flatten(), aa.flatten()))
+    #xydata = np.array((xx.flatten(), yy.flatten()))
+    zs = np.array([f(xy, 1.5,  2) for xy in xydata.T])
+    zdata = zs.reshape(xx.shape)
```
tBuLi: This reshaping to xydata is no longer required, you can just use
```python
    fit = Fit(model, x=xx, y=yy, ...)
```
In this case. I think that in order to generate your data here it might even be easiest to just do model(x=xx, y=yy, a=1.5, b=2.0)

tools/interactive_fit.py
```diff
@@ -0,0 +1,359 @@
...
+#    interp2 = LinearNDInterpolator(xydata.T, zdata.flatten())
+#    print(interp2(0.5, 1, 0.2))
+
+    x = Variable()
+    y = Variable()
+    a = Variable()
+    b = Parameter()
+
+    model = cos(a*x) * b*sin(y)
+    model = f_sym(x, y, a, b)
```
tBuLi: this line is very frowned upon. Right now, as we speak.

tools/interactive_fit.py
```diff
@@ -0,0 +1,359 @@
...
+    fit = InteractiveFit3D(model, xydata.T, zdata.flatten())
```
tBuLi: So if I did everything correctly, this should now be simply
```python
    InteractiveFit3D(model, xx, yy, zdata)
```
Although I would prefer even in such a simple case to use keyword arguments for readability:
```python
    InteractiveFit3D(model, x=xx, y=yy, z=zdata)
```

Pitje06:
Updated InteractiveFit2D to reflect (most of) tBuLi's comments. 
InteractiveFit3D is considered broken, and will be reworked to implement Marginal Distributions (https://en.wikipedia.org/wiki/Marginal_distribution).

As for tests... I wouldn't even know where to start (A). But that may change of the course of a few days.

Pitje06:
So, to reduce the dimenionality in the data; some more ideas.
Two datasets need their dimensionality reduced: 1) the original dataset (scatterplot) and 2) the function drawn with the current guess.

1) given x_hat  and y_hat as independent datapoints and z_hat as corresponding dependent data one can construct the probability of finding x and y p(x, y) by doing a kernel density estimation.
Next, \<x_hat\> needs to be made by binning x_hat. \<z_hat\> can then be made by averaging all x_hat, y_hat in each \<x_hat\> using p(x_hat, y_hat) as weights. Note that this needs to be normalized by sum(p(x_hat, y_hat)) for all x_hat, y_hat in \<x_hat\>.

2) Make a grid of x, y points, and evaluate your guessed function f(x, y) to find z. For plotting, calculate \<z\> for each x by averaging z, using p(x, y) as weights. Note that this needs to be normalized by p(x).

@tBuLi: Is it clear what I mean by this? And do you agree?

tBuLi:
Ok I looked into the math. For point 2 we agree, although I will show the maths I did to make sure we're doing the same thing.

For the normalization we do:
A= \int  \int z(x, y) dx dy

Where the integrals go over the domain we are interested in. Then:

p(x, y)= \frac{z(x, y)}{A}

Gives the pdf of z. In order to plot z vs x, we can now plot the following:

E[z](x) = \int z(x, y) p_Y(y) dy \\p_Y(y) = \int p(x, y) dx

Where indeed this is <z(x)> after averaging out y.

For point 1, the dataset, I think we could try repeating the procedure, replacing p(x, y) with the pdf of the kernel density. All that would be left is to figure out the conversion factor of the data back to the ámplitude' of the data. I'll think about that some more in the morning ;).

symfit/contrib/interactive_fit/examples/vector_valued_2D.py
```diff
@@ -0,0 +1,29 @@
...
+import numpy as np
+
+
+x = Variable()
+y1 = Variable()
+y2 = Variable()
+k = Parameter(900)
+x0 = Parameter(1.5)
+
+model = {y1: k * (x-x0)**2,
```
tBuLi: I'd prefer if this was spread across multiple lines:

```python
model = {
    y1: k * (x-x0)**2,
    y2: x - x0
}

File: symfit/contrib/interactive_fit/interactive_fit.py [hunk 16]
```diff
@@ -0,0 +1,194 @@
...
+#
+#if parse_version(SUPPORTED_VERSION) != parse_version(SYMFIT_VERSION):
+#    raise EnvironmentError("Your symfit version is not supported. YMMV.")
+
+
+class InteractiveFit2D(Fit):
+    """A class that provides a visual_guess method which provides
+    an graphical, interactive way of guessing initial fitting parameters."""
+
+    def __init__(self, *args, **kwargs):
```
tBuLi: Can this be split up into multiple methods? This function is way to long, making it hard to read.

symfit/contrib/interactive_fit/interactive_fit.py
```diff
@@ -0,0 +1,194 @@
...
+        if 'n_points' in kwargs:
```
tBuLi: Use the kwonly decorator from symfit.core.support

symfit/contrib/interactive_fit/interactive_fit.py
```diff
@@ -0,0 +1,194 @@
...
+        independent_var : Variable
+            The independent variable to calculate the data for
+
+        Returns
+        -------
+        x_points, y_points
+        """
+        x_points = self._x_points[independent_var.name]
+        arguments = {independent_var.name: x_points}
+        arguments.update({p.name: p.value for p in self.model.params})
```
tBuLi: Because this happens so often, I build a nice method in support called ``key2str``. Check it out! It means you can build your dict with any type of object that has __str__ defined as keys, and upon calling you just do **key2str(my_dict)

Pitje06:
I can't take keywords only, I expect to find an model as positional argument.
key2str is not applicable, since self.model.params is a list of parameters.
I'll look at splitting the init method later, if you think it's really nescessary - I might split off setting up the sliders to a separate method, I don't really see any gain in splitting off the rest as well.

tBuLi:
On keywords only, your problem is that what you would like to do:
```python
def __init__(self, n_points, *args, **kwargs)
```
but this disrupts the assignment of the model argument, since that is the first one in the parent class. Right? 

This is kind of a downside to py2 I think, because the standard way to do it would be to break DRY, which is not only silly but could also introduce bugs. So your current solution is beter; pop it from kwargs. The downside to this method is that you cannot read from the function declaration which kwargs are accepted by the method. On top of that, this way you already made it a kwonly argument for all intents and purposes.

So to deal with such issues is precisely what the ``@keywordonly`` decorator is for.
```python
@keywordonly(n_points=100)
def __init__(self, *args, **kwargs)
    n_points = kwargs.pop('n_points')
```
This is readable from the function declaration, because the information on which kwargs there are is where you need it to be. And the decorator guarantees that the pop works.

key2str is a nice clean up developed for situation such as yours, because rather than doing
```python
a = {p.name: p.value for p in self.model.params}
f(**a)
```
you can do
```python
a = {p: p.value for p in self.model.params}
f(**key2str(a))
```
This is not a really big improvement I agree. I just mentioned this as a tip of a way it could be done, but the way you currently did it is off course equivalent. The reason for pointing it out is because in #41 we talked about the issue of when things should have a str as key and when an object. I think we should strive towards objects always, and when you want to call a function with such a dict use ``key2str``. This makes for a clean internal API, I think you'll agree.

As to cleaning up ``__init__``, yes I could have saved some confusing by saying that the setup of all the axis etc is precisely what bugs me. For the rest I agree that those thing belong in init. But setting up axis and figures involves so much boilerplate code that it obfuscates the logic of your init method.

I apologize for being such a pain in the neck about this, and I agree that since this is noted as contrib maybe different rules apply. But at the same time I do want to make sure that everything is as consistent as possible with the core API to make sure this code keeps working in the future.

Pitje06:
It's good to have someone else look at my code for a change :) 
The latest commit should address all your suggestions ;)
I think the `keywordonly` decorator is rather elegant, I should've read it's docs...
For future compatibility I also included the key2str function where possible, so that when you fix #41, you can make key2str an no-op.

Pitje06:
Ok, so after some discussion we agreed that guessing initial values is not fitting. This means that I should  not inherit from `symfit.Fit`, but rather, from something that does just the handling of data, variables and parameters.  Let me know what it's called ;)
Did I forget anything?

tBuLi:
I was think of naming that baseclass ```HasData```, what do you think? That can than be used as a base for anything.

Unfortunately I'm drowning in other obligations right now so I can't make this on short notice. 

Could I ask you to do it? All you need to do is isolate ```__init__``` on ```BaseFit``` and everything that is refered to in this init to do with the data handling. If I did my job correctly it shouldn’t be that obscure.

Pitje06:
I disagree with the name, `HasData` sounds like a boolean function (`if HasData(model): ...`). I don't know what it should be though.
And I'll take a look :)

Pitje06:
Looking at BaseFit, I would just copy the entire thing except `execute`, `error_func` and `eval_jacobian`. Am I missing something?
As for a name: `DataContainer`, `DataObject`, ...?

tBuLi:
I agree that ```HasData``` sounds like a boolean. However, I also don't like ```DataContainer``` because these objects will not be containers. They are something that takes data. So how about ```TakesData```?

I think that is very clear:
```python
class BaseFit(TakesData):
    pass
```

Pitje06:
Still not the best name I think, since there's no such thing as "a TakesData", while "a DataContainer" does make sense, but meh. I'll make it so.

Pitje06:
There, that should do it. I get two failing tests, but I don't think it's related (test_nonlinearfit and test_likelihood_fitting_gaussian give wrong numbers).
It might be worthwhile to split fit.py in smaller pieces, it's rather large IMHO, but that's up to you.

Pitje06:
Ok, if I did all this correctly (which is not very probable), PR #55 creates TakesData from BaseFit; and this branch should use those changes.

Pitje06:
Further musings: should I add an plotlegend? If so, what's the way to get the symbolic expression of model, preferably in LaTeX form?

tBuLi:
This should work out of the box:
http://docs.sympy.org/dev/modules/printing.html#sympy.printing.latex.latex
You can't call it on model directly, but you can on its components

Pitje06:
There is an issue with this as it is now, since you (apparently) need to list all directories you want to include in setup.py.
I added contrib and contrib.interactiveguess for now, but I think all subdirectories of contrib that contain a __init__.py file should be included automagically. See also: setuptools.find_packages()

Pitje06:
This branch requires matplotlib for its tests, but this is of course not included in the requirements. How do you want to proceed with Travis on this branch?

Pitje06:
I found an appealing solution to the dependency problem. I added an [extras] section to setup.cfg, such that users can now do `pip install symfit[contrib]` to also install all dependencies required by contrib packages.
This is also reflected in .travis.yml.

symfit/contrib/interactive_guess/interactive_guess.py
```diff
@@ -0,0 +1,182 @@
...
+        ncols = int(np.ceil(len(self._projections)/nrows))
+
+        # Make all the subplots: set the x and y limits, scatter the data, and
+        # plot the putative function.
+        self._plots = {}
+        data = self._get_data()
+
+        for plotnr, proj in enumerate(self._projections, 1):
+            x, y = proj
+            if hasattr(self.model, 'dependent_derivatives') and Derivative(y, x) in self.model:
```
tBuLi: I think `Derivative(y, x) in self.model` will suffice, right?

symfit/contrib/interactive_guess/interactive_guess.py
```diff
@@ -0,0 +1,182 @@
...
+        ncols = int(np.ceil(len(self._projections)/nrows))
+
+        # Make all the subplots: set the x and y limits, scatter the data, and
+        # plot the putative function.
+        self._plots = {}
+        evaluated_model = self._eval_model()
+
+        for plotnr, proj in enumerate(self._projections, 1):
+            x, y = proj
+            if hasattr(self.model, 'dependent_derivatives') and Derivative(y, x) in self.model:
```
tBuLi: I think `Derivative(y, x) in self.model` suffices, right?

Pitje06:
Thanks for the review. bae9f39 is ok if and only if Fit will always be a local optimization method. I have no problems with the rest of the changes. I'll process the rest of your comments later today.

Pitje06:
Commit message ecdb686 is wrong. Should be #107

symfit/contrib/interactive_guess/examples/ODE_example.py
```diff
@@ -0,0 +1,29 @@
...
+                          0.0324, 0.026])
+
+# Define our ODE model
+A, t = variables('A, t')
+k = Parameter()
+model = ODEModel({D(A, t): - k * A}, initial={t: tdata[0], A: concentration[0]})
+
+guess = InteractiveGuess2D(model, A=concentration, t=tdata, n_points=250)
+
+print("Guessed values: ")
```
tBuLi: Actually, I think InteractiveGuess objects should have `__str__` defined such that when you call print(guess), we get essentially this report.

Issue #50, opened 2016-02-09
test_gaussian_2d_fitting fails


tBuLi:
This test has been unreliable since day 1. For one, I have noticed different behavior between py2.7 and py3.x. This seems to be related to x, y or y, x ordering in 2D numpy arrays. 

I therefore ignore it when this fit errors, since I assume it not to be fundamental to ``symfit``. Obviously assumption is the root of all cockups, so it would be nice if someone could investigate.

Pitje06:
So, after looking at this for a small while, it appears to be a bit of an Heissenbug.

If I add an `np.random.seed(0)`, the test fails in both python 2 and 3. With the seed `999` it succeeds in both versions.

So in short, your data isn't good enough. But that is kind of funny in it's own right. Creating 100 times as much data makes it work (and wrecks your memory). An alternative fix is initializing `A = Parameter(800)`.

It sort of makes sense: 1D Gaussians are already difficult to fit to, due to a lot of local minima, let alone 2D Gaussians. This is one of the reasons visual guessing your initial parameters can be so valuable.

You should have a pull request somewhere soon-ish.

PR #51, opened 2016-02-10
Resolved issue #50

Pitje06:
Resolved by adding a setup method that initalizes the random seed, and gives an initial guess for A.

Further, did some pep8 cleaning, but it's still a mess.


tBuLi:
Thanks for looking into this! 

I just tried to merge this into main but I ran into a lot of merge issues. I suspect you were working on a quite outdated version of symfit so I gave up on trying to fix them :P. Could you please pull the latest version and start from there? That will make merging this fix a lot easier.

PR #55, opened 2016-04-06
Create takes data

Pitje06:
Split off TakesData from BaseFit, as discussed in PR #33.


symfit/core/fit.py
```diff
@@ -810,14 +810,9 @@ def __init__(self, model, *ordered_data, **named_data):
...
             # print(var, sigma)
-            try:
-                iter(self.data[sigma.name])
-            except TypeError:
-                try:
-                    self.data[sigma.name] *= np.ones(self.data[var.name].shape)
-                except AttributeError:
-                    # If no attribute shape exists, data is also not an array
-                    pass
+            if not isinstance(self.data[sigma.name], Iterable) and hasattr(self.data[var.name], 'shape'):
```
tBuLi: Do you know if the explicit checking for isinstance breaks duck-typing? What if someone makes an iterable object without inheriting from that ABC? I think that should be allowed. I do agree with the hasattr.

PR #74, opened 2016-10-18
Fix TakesData

Pitje06:
I left an `isinstance` test in the previous PR, and removed Iterable from the imports. This broke.
This PR rewrites that case to EAFP also.


tBuLi:
How weird, that change was not highlighted in my github. Well anyway, it looks good now, I'll merge it.

Issue #73, opened 2016-10-17
Travis-CI


tBuLi:
This looks pretty amazing. Apparently this app automatically runs tests upon pushing a change, and can also auto compile pip wheels. If true this could remove a lot of the manual labor I've been doing.

https://travis-ci.org

Pitje06:
I've seen it been used for some larger projects, I think it should just work (TM).

Pitje06:
@jbarnoud has experience with Travis, and is willing to look at it. I suggest assigning this to him :)

jbarnoud:
It is on my TODO list.

tBuLi:
I could use an extra pair of eyes on my `deploy-fix` branch. It seems to be very close to auto-deploying on testpypi.python.org, except for a rather annoying bug in the [travis build](https://travis-ci.org/tBuLi/symfit/jobs/175760177):
```
line 491: sh: 1: twine: not found
```
Even though it should be auto installed and appears to have been in the build log. I therefore added the lines 
```
before_deploy:
  pip install twine
```
as suggested [here](https://github.com/travis-ci/dpl/issues/299) to make sure but that did not solve it. (And that issue seems closed with dpl 1.7.19 and we get 1.8.24 by now.) 

Any ideas? I could still try `--user` on the install or `edge: true`, though neither should be necessary anymore according to [299](https://github.com/travis-ci/dpl/issues/299). I am hesitant to keep making minor fixes on the branch but I can't run travis locally on my very old mac. Does any of you see something that I missed?

Btw, I purposely did not make a pull request for this branch, because travis skips the deploy step for PRs. So unfortunately that is not an option in this case.

p.s. it tries to deploy to test for any commit, and to live pypi only for a tagged commit. So line 508 in the build log doesn't mean anything for the deploy to test.

jbarnoud:
Maybe $PATH needs to be updated?

Pitje06:
https://github.com/travis-ci/dpl/issues/296
Maybe try putting the `pip install twine` in the install section.

Pitje06:
As an aside, I'm also in favor of moving to pip instead of conda (because that does magic travis caching), but https://github.com/travis-ci/travis-ci/issues/2650 might be an issue.

tBuLi:
Putting `pip install twine` in the `install` section made [no difference](https://travis-ci.org/tBuLi/symfit/jobs/176377625). 

I am not comfortable enough with $PATH to know how to approach that.

[...]

PR #76, opened 2016-10-23
Add a config file for travis-ci

jbarnoud:
This PR allows to activate continuous integration using Travis-CI, see #73. To activate Travis, and once the file is merged, it is still needed for @tBuLi to create an account on Travis and to turn the continuous integration for the symfit repository.

The `.travis.yml` file gives the build instruction to Travis. It installs the dependencies using conda, and run the tests using nose.

At the moment, the tests fail. But they fails in the same way on my computer.

The results of the test can be seen on https://travis-ci.org/jbarnoud/symfit

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,34 @@
+language: python
+# Choose the supported versions of python.
+python:
+  - "2.7"
+  - "3.3"
+  - "3.4"
+  - "3.5"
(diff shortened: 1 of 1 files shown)


tBuLi:
Thanks a lot for looking into this @jbarnoud! 

I will look into creating an account as soon as possible. And the tests fail in an expected manner but I will change them to reflect that ;).

Issue #79, opened 2016-10-24
Tests depend on matplotlib/seaborn


Pitje06:
tests/tests_ode.py imports both seaborn and matplotlib, but doesn't use them. They should be removed from the test, because they pollute the dependencies (also for Travis, #76).

Pitje06:
This is fixed, and I'm closing this now.

PR #80, opened 2016-10-24
Clean tests

Pitje06:
This is a small start for the cleaning of the tests. I suggest we work on this branch to streamline things a bit. tests_ode.py should be sane now. 
What testing framework do you use for running tests? Nose or PyTest? I was just told nose is no longer maintained, so I think I prefer pytest, but you pick.


Pitje06:
Note that commit 8a6c421 fixes #79

symfit/tests/tests_ode.py
```diff
@@ -105,8 +99,8 @@ def test_simple_kinetics(self):
         a0 = 54 * 10e-4
 
         model_dict = {
-            D(a, t): - k * a**2,
-            D(b, t): k * a**2,
+            Derivative(a, t): - k * a**2,
```
Pitje06: Your namespacing is ~~a mess~~ complicated ;)

from symfit import D:
ImportError: cannot import name 'D'

from symfit.api import D:
ImportError: cannot import name 'D'

from symfit.support import D:
ImportError: No module named 'symfit.support'

I finally found the answer:
from symfit.core.support import D
But renaming D to Derivative was the faster solution. Do you actively dislike Derivative?

tBuLi: Yes, I moved its definition from `api` to `support` in a recent commit, where it belongs. But apparently I forgot to add the line `from symfit.core.support import D` to `.api` apparently, that is the real issue here.

I don't dislike Derivative perse for internal usage, but I feel like tests are an external script to the actual package. As such, they should use the public api as documented, and preferably not internal functions. Of course that is not always possible, but as a general philosophy I try to stick to it.

tBuLi:
44e7b73 was long overdue, thanks ;)

symfit/tests/tests_ode.py
```diff
@@ -137,17 +137,19 @@ def test_single_eval(self):
...
         harmonic_model = ODEModel(harmonic_dict, initial={t: 0.0, x: 1.0, y: 0.0})
 
         tdata = np.linspace(0, 100, 101)
         X, Y = harmonic_model(t=tdata, k=0.1)
         for index, t in enumerate(tdata):
             X_point, Y_point = harmonic_model(t=t, k=0.1)
-            self.assertAlmostEqual(X_point, X[index])
-            self.assertAlmostEqual(Y_point, Y[index])
+            np.testing.assert_allclose(X_point, X[index])
+            np.testing.assert_allclose(Y_point, Y[index])
```
tBuLi: Why is this preferred? As it stands now `X_point, Y_point` are just floats right? Maybe you also mean to get rid of the for loop?

Pitje06:
I noticed quite some tests 1) don't have docstrings and 2) don't assert anything. I would be happy to do something about this, but I can't, since I have no idea what they're supposed to do...

Issue #75, opened 2016-10-18
Global optimization


pckroon:
Fitting is effectively a minimization problem, where you minimize chi-squared. Usually this is done using a local minimization algorithm such as steepest descent. 
This means the answer you get out of your fit is dependant on your initial guesses for your parameters. This is what inspired PR #33. 
As an alternative, it may be beneficial to either do the entire fit using a global optimization algorithm such as Differential evolution or Basinhopping. If this is too poor performance wise, one of these may be used to give a coarse guess for the initial values, which are then polished using "normal" local minimization algorithms.
Since these algorithms do not use the jacobian (unfortunately), they may be exceptionally suited for ODE problems.
EDIT: Actually, basinhopping can use a gradient based local optimization method between hops, but I don't know how well the algorithm is suited for fitting.

See also: https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.differential_evolution.html#scipy.optimize.differential_evolution
https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.basinhopping.html#scipy.optimize.basinhopping
http://ab-initio.mit.edu/wiki/index.php/NLopt_Algorithms (Find alternative implementations, or reimplement yourself, since you can't pip install this beast)

PR #86, opened 2016-10-31
Global fit

pckroon:
This adds the option to do a global parameter optimization to minimize chi^2 (instead of a local one) using differential evolution.
The default DE metaparameters still need some tweaking for the sake of efficiency though.
Fixes/implements #75


pckroon:
Not compatible with #133; closed.

Issue #69, opened 2016-10-13
'Accessing the Results' is still old style


tBuLi:
The accessing the results part of the docs should be updated to reflect the new syntax. #43

PR #93, opened 2016-11-29
Better documentation

### LINKED ISSUE #77: Oscilatory ODE fitting example for docs
Oscilatory ODEs are quite common, and it might be good to add an example of that to the docs, or to the tests at the least.

Example:
http://stackoverflow.com/questions/34109050/scipy-optimize-leastsq-dimension-error-fitting-ode

Pitje06:
A start to brush up the documentation. The first commit fixes #77, but there's probably more to come :)

PS. Can we exclude this PR from travis?


docs/fitting_types.rst
```diff
@@ -256,6 +256,37 @@ Let's plot the model for some kinetics constants::
...
+More common examples, such as dampened harmonic oscillators also work as expected:
+
+  # oscillator strength
+  k = symfit.Parameter()
+  # mass, just there for the physics
+  m = symfit.Parameter(fixed=True)
+  # Dampening factor
+  gamma = symfit.Parameter(1)
+  
+  x, v, a, t = symfit.variables('x, v, a, t')
```
tBuLi: `a` is redundant, since `a` will be made into an expression in the next line.

docs/fitting_types.rst
```diff
@@ -256,6 +256,37 @@ Let's plot the model for some kinetics constants::
...
+  
+  a = (-k * x - gamma * v)/m
+  ode_model = symfit.ODEModel({
```
tBuLi: I would prefer to define the `model_dict` separately _and_ in the same style as found elsewhere in the docs.

docs/fitting_types.rst
```diff
@@ -256,6 +256,37 @@ Let's plot the model for some kinetics constants::
    :width: 300px
    :alt: ODE integration
 
+More common examples, such as dampened harmonic oscillators also work as expected:
+
+  # oscillator strength
+  k = symfit.Parameter()
```
tBuLi: `k = Parameter()`

docs/fitting_types.rst
```diff
@@ -256,6 +256,37 @@ Let's plot the model for some kinetics constants::
...
+  # mass, just there for the physics
+  m = symfit.Parameter(fixed=True)
```
tBuLi: If mass is not something that was to be fitted but rather a constant, we can just leave it as `m = 1` for example. I think that's clearer.

Pitje06:
For completeness' sake, a fix for #69 is in there somewhere as well.

docs/tutorial.rst
```diff
@@ -121,22 +125,26 @@ There is a preferred way to resolve this. If any kind of fitting object has been
...
     model = {y_1: a * x**2, y_2: 2 * x * b}
-    fit = Fit(model, x=xdata)
+    fit = Fit(model, x=xdata, y_1=y_data1, y_2=y_data2)
     fit_result = fit.execute()
 
     y_1, y_2 = fit.model(x=xdata, **fit_result.params)
 
-This returns a tuple with the components evaluated so through the magic of tuple unpacking``y_1`` and ``y_2`` contain the
-evaluated fit. Nice!
+This returns a named tuple with the components evaluated so through the magic of tuple unpackingi ``y_1`` and ``y_2`` contain the
```
tBuLi: named tuple -> \`namedtuple\`. unpackingi -> unpacking.

docs/tutorial.rst
```diff
@@ -121,22 +125,26 @@ There is a preferred way to resolve this. If any kind of fitting object has been
...
+evaluated fit. Nice! It is usually beter to do the unpacking explicitely though.
```
tBuLi: Disagree with that last statement. First line of the style guide:

> Anything Raymond Hettinger says wins the argument until I have time to write a proper style guide.

Joking aside, in simple cases such as this, tuple unpacking is just so much clearer and easier. And alphabetic ordering of the dependent variables is guaranteed so no ambiguity there.

docs/tutorial.rst
```diff
@@ -121,22 +125,26 @@ There is a preferred way to resolve this. If any kind of fitting object has been
...
 
     from symfit import Model
 
     model_dict = {y_1: a * x**2, y_2: 2 * x * b}
-    model = Model.from_dict(model_dict)
+    model = Model(model_dict)
+
+    outcome = fit.model(x=xdata, a=2.4, b=0.1)
+    y_1 = outcome.y_1
+    y_2 = outcome.y_2
```
tBuLi: See previous statement. The tuple unpacking you replaced is so much clearer.

docs/tutorial.rst
```diff
@@ -121,22 +125,26 @@ There is a preferred way to resolve this. If any kind of fitting object has been
...
 
-    y_1, y_2 = fit.model(x=xdata, a=2.4, b=0.1)
+.. examples above overwrites the variables y_1 and y_2. Plus, the parameters and variables are not defined anywhere...
```
tBuLi: yeah the overwriting could be changed perhaps. This line is actually only meant to show how to make a `Model` from a dict. Could perhaps be turned into a fully working example, but I feel that this far into the docs a user should know how to do that right?

docs/tutorial.rst
Pitje06: I agree with myself (A).
Explicit is better than implicit. And I think it would be good to show both methods. I changed the wording to be less strong.

docs/tutorial.rst
```diff
@@ -34,40 +34,41 @@ Printing ``fit_result`` will give a full report on the values for every paramete
...
 
-In the example above, we might change our ``Parameter``'s to the following after looking at a plot of the data::
+In the example above, we might change our :class:`~symfit.core.argument.Parameter`'s to the following after looking at a plot of the data::
 
   k = Parameter(value=4, min=3, max=6)
 
-  l, m = parameters('b, c')
-  l.value = 60
-  l.fixed = True
+  a, b = parameters('b, c')
```
tBuLi: Problem persists ;).

docs/tutorial.rst
```diff
@@ -110,37 +111,40 @@ Let's try our luck with a bivariate normal distribution::
...
 
-    model_dict = {y_1: a * x**2, y_2: 2 * x * b}
-    model = Model.from_dict(model_dict)
+    outcome = model(x=xdata, a=2.4, b=0.1)
+    y_1_result = outcome.y_1
+    y_2_result = outcome.y_2
 
-    y_1, y_2 = fit.model(x=xdata, a=2.4, b=0.1)
 
 symfit exposes sympy.api
```
tBuLi: sympy.api -> sympy since sympy.api does not actually exist. Mistake from working with a lot of Enthought projects that did follow that paradigm at the time.

PR #97, opened 2016-12-07
Documentation improvements

### LINKED ISSUE #96: Multivariate example
There is currently no clear example of fitting to multivariate data in the docs, and it would clarify a lot if e.g. the grid fitting test could be worked into an example.

pckroon:
Just fixed some small typos for now. 
Leave it open so I can fix #96 sometime later this week.


tBuLi:
Since tBuli/symfit#104, the docs will also be automatically tested. However, I did end up telling it to ignore errors caused by referring to `scipy` and `symfit` directly with the `:mod:` directive. If we have any bright ideas on why, we could update that in this PR.

pckroon:
Very much superseded.

PR #104, opened 2016-12-12
Added sphinx-build script to autotest doc building

tBuLi:
Test building of the docs.


Pitje06:
Also, make sure you build the docs just once, rather than once for all Python versions. (And of course, d9ef052 isn't going to help if  you have the lines commented ;)

tBuLi:
Good point. From testing locally I found that with the new keywordonly function, the signature is read beautifully by sphinx in 3.5, but ignored in 2.7 because it still uses getargspec. So it's easy to choose where to build our docs.

PR #107, opened 2016-12-15
Fixed an issue with the dimensionaly of the Dfun argument to odeint

tBuLi:
In checking interactive_guess I found that it's example with 2D ODE's didn't work. It's weird that similar examples in the tests did run properly, but the jacobian was indeed a vector and not a matrix as it should be.

This branch fixes this, and I will also use this later to implement `numerical_jacobian` on ODEModels later.

### DIFF
--- a/symfit/core/fit.py
+++ b/symfit/core/fit.py
@@ -2186,8 +2186,10 @@ class ODEModel(CallableModel):
     @property
     @cache
     def _njacobian(self):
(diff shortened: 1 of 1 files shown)


tBuLi:
@Pitje06: any ideas as to why this could be? I can't reproduce this problem from my own code, so something happens in interactive_fit which forces this bug to manifest.

Pitje06:
No, I don't know why this happens. I worked around it by changing my example from
```
{
A: k * A, 
B: -k * A
}
```
to
```
A: k * A**2, 
B: -k * A**2
```

Issue #91, opened 2016-11-29
FitResults should no longer have private parts.


tBuLi:
FitResults uses `private` attributes to make sure users can't modify fit results. For example, `FitResults.__iterations` contains the number of iterations needed for the fit. This can be accessed on a read-only basis via the `FitResults.iterations` property.

In order to make the internal API cleaner, I propose changing all double underscores to single underscores. The public API will still be to use the read-only property as a save guard. However, internal modification of an instance of `FitResults` before returning it to the user will be much easier this way.

The same thing goes for the ParameterDict, though that should be removed entirely shortly and is therefore not worth the time. (#57, #45)

Issue #57, opened 2016-05-17
ParameterDict has an identity crisis


Jhsmit:
Is it a bird? Is it a list? Is it a dict? Is it superman? 
It seems to be a list/dict hybrid, which might be intentional.

```
for key in result.params:
    result.params[key]
```

In any case, this does not work. What is the recommended method for getting unknown parameter values and their names out of this monster?

Jhsmit:
PS. of course I did read the documentation and docstrings. 

```
for par in result.params:
    name = par.name
    value = par.value
```
This gives me incorrect values for value, perhaps it gives the initial guess?

tBuLi:
Yeah it made sense at the time. I opened a discussion about it a while ago: #45.

The reason you can't iterate over it directly is because iterating over the ParameterDict means you will iterate over the Parameter objects which are indeed your initial guesses.

To iterate over the fitted value's I would do:
```python
for param in fit_result.params:
    value = fit_result.value(param)
```

Any suggestions on the future of this are welcome in #45.

Issue #45, opened 2015-12-21
Two different syntaxes for getting Parameter values?


tBuLi:
Currently there are two different ways to get the value and stdev of a parameter:

``` python
# a = Parameter()

assert fit_result.params.a == fit_result.value(a)
```

This kinda conflicts with the principle that there should be only one obvious way to do something. Which is better? 

If we decide to remove the first one, that means the ParameterDict object can be removed entirely and replaced with an OrderedDict. This is something I would like, as the current ParameterDict is kind of a "don't touch it, it works"-thing.

Furthermore, the second syntax is better when dealing with dynamically generated `Parameter` objects whose name you don't know or care about.

The only thing going for the first one is that I think it's rather elegant. 
@Jhsmit, @Pitje06, @Eljee What do you think?

Pitje06:
I always find these questions very hard... But here's my two cents.

I'm in favor of the first syntax, since IMHO it's more Pythonic (attributes over getter functions). Optionally you could change it to `fit_result.a`. 
The main downside of the first variant is however, that you need to know the names of your parameters when you're writing your code. Otherwise you're stuck with getattr to retrieve your values (but this should be <20% of all use cases).
Also, `a` should be a parameter object (subclass from float) that has an stddev attribute so it can be used as transparently as possible. If possible, it should be tied in with one of the sympy primitives such that errors get propagated.
As for the implementation, I don't see why you can not implement this relatively simple with a `__getattr__` method that first tries to get it from the object itself (`__dict__`?) and otherwise from a param_dict attribute.

tBuLi:
Thanks for your reply. Upon reading it I realized I should have thought about my original post longer because I didn't specify all the problems I have with the current ``fit_result.params.a`` syntax. So here goes:

To get standard deviation in either paradigm you do
```python
assert fit_result.params.a_stdev == fit_result.stdev(a)
```
(Btw, the first works by overloading ``__getattr__`` like you suggest)

However, this way of accesing the value and stdev has no consistent way of getting covariances between fit parameters. Therefore it would be nice to have one syntax which feels self-consistent. This is why I implemented the following functions (#43):

```python
fit_result.value(a)
fit_result.stdev(a)
fit_result.covariance(a, b)
fit_result.variance(a) 
```

They all have a similar feel to them, which is why I like them better.

Now, the ``ParameterDict`` was implemented purely to create the ** shorthand when calling models:
```python
y = model(x=xdata, **fit_result.params)
```
[...]

Issue #85, opened 2016-10-28
FitResults should show more goodness of fit qualifiers besides R^2


tBuLi:
More goodness of fit qualifiers should be added to `FitResults` besides R^2. Depending on the fitting type, we also directly know chi_squared or log_likelihood. These should therefore be available from `FitResults`.

So at the very least `chi_squared`, `degrees_of_freedom` (or `dof`) and `chi_squared_reduced`, `likelihood` and `log_likelihood` can be added. In case one of these is not present because it is not sensible to ask, it should return `None`.

`likelihood` should be a property, `log_likelihood` an attribute. `likelihood` can then be calculated from log_likelihood.

`chi_squared` and `degrees_of_freedom` should be an attribute, `chi_squared_reduced` is given by dividing `chi_squared` by `degrees_of_freedom`.

Jhsmit:
Are GoF qualifiers available in FitResults yet? I'm interested in the ``chi_squared``, which I currently get by: 
```python
@property
def chi_squared(self):
    return np.sum(
        self.fit.model.numerical_chi_squared(
            *self.fit.fit.data.values(), **self.result.params
        )
   )
```

Where I've added the second `fit` in `fit.fit.data` when updating to symfit 0.3.5. Is this still the correct/best way to get the ``chi_squared``?

tBuLi:
Yeah, unfortunately you have to create a workaround like yours. So you created this property on an object you use? Personally I have done the same as you, but it should also be possible to use
```python
 np.sum(result.infodict['fvec']**2)
```
``result.infodict['fvec']`` contains theresiduals evaluated at the solution as it is returned by ``scipy.optimize.leastsq``. As I am typing this I realize there might be a bug here:  I think ``fvec`` might be a vector of the resuduals when returned under the hood by NumericalLeastSquares and the scalar chi^2 when ConstrainedNumericalLeastSquares was used. So please check this against your method before implementing this :P.

Jhsmit:
Yes I should have mentioned that this a property in one of my own objects. And your suggestion indeed gives the same result as my property. I'm not using constraints at the moment so I should be fine.

tBuLi:
Glad to hear it. Could you briefly try to use a `ConstrainedNumericalLeastSquares` object instead of the `Fit` object? If that is not the same it is definitely a bug.

PR #111, opened 2017-02-05
Fit results cleanup

tBuLi:
This PR aims to solve #57, #45, #91.

It makes a rethink of the ``**fit_result.params`` syntax necessary, to prevent confusion with ``Model.params`` which is entirely different. (The first should be an `OrderedDict` with the optimized parameters, second a list of `Parameter` objects.)

Options are ``**fit_result`` or a renaming of the dict to e.g. ``**fit_results.fit_params``.

Because this will be a significant API change in either case, a good rethink is needed.

I am against the first option for the same reasons I did not originally do this: if you iterate over a FitResults object, what do you expect to iterate over? I can only see two options: it's entire contents, or different solutions to the fitting problem (local minima). But not just the fit parameters. Additionally, making a custom mapping object adds complications over just using the build-in ones (I learned that the hard way, see #57).

So my preferred solution is definitely to just give ``fit_result.params`` a different name, and have it be an `OrderedDict`. 

Suggestions are welcome.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -104,9 +104,8 @@ fit result::
 
   y = model(x=xdata, **fit_result.params)
   
-This unpacks the :attr:`~symfit.core.fit.FitResults.params` object as a
-:class:`dict`. For more info view :class:`~symfit.core.fit.ParameterDict` in
-the :ref:`apidocs`.
+This immediately unpacks an :class:`~collections.OrderedDict` containing the optimized fit
+parameters.
 
 Named Models
 ------------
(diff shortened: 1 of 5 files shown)


tBuLi:
#85 is now also included in the scope of this PR

tBuLi:
Comming back to the discussion on what to do with the `**fit_result.params` syntax, I have now at least prevented all internal confusion within the FitResults object by initiating it with a model, rather than a list of Parameters. 

So no API change is strictly necessary anymore, but probably it is still the right thing to do. I would there propose to depricate `**fit_result.params`  and inform people to move to  `**fit_result.param_dict` .

PR #112, opened 2017-03-16
Matplotlib travis

### LINKED ISSUE #109: Matplotlib beta in travis.yaml
Matplotlib 2.0.0 has been released, so travis should install that rather than the beta version.

pckroon:
2 small commits; the first to tell travis to no longer use MPL beta, and the second to reflect my new github username.
Fixes #109

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ install:
   - pip install --upgrade pip setuptools
   - pip install --upgrade pytest
   - pip install --upgrade sphinx
-  - pip install --only-binary numpy,scipy,matplotlib==2.0.0rc2 .[all]
+  - pip install --only-binary numpy,scipy,matplotlib .[all]
   - echo $PATH
 # Run tests
 script:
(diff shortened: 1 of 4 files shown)


pckroon:
I think the failing tests come from a change in how scipy's curve_fit deals with sigma in version 0.19.0.
https://docs.scipy.org/doc/scipy-0.18.1/reference/generated/scipy.optimize.curve_fit.html
https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.curve_fit.html
What the tests give as sigma is sigma = array(31.4), array(10.0) and array(0.0031). Which was always incorrect, since it is not an M-length sequence.

tBuLi:
Yes, it seems like it was allowed to be a scalar in the past, even though the docs always said it should have been an M-length array. So changing 
```python
sigma = 10.0
```
to 
```python
sigma = 10.0 * np.ones(N)
```
should solve the problem in the tests, right?

PR #119, opened 2017-04-10
fixed fit by specifying seed for gaussian

Jhsmit:
Added a fixed seed to gaussian example so fitting always works, fixes #119

### DIFF
--- a/examples/gaussian.py
+++ b/examples/gaussian.py
@@ -15,6 +15,7 @@ x0 = Parameter(name='x0', value=15.0, min=0.0)
 model = A*exp(-((x - x0)**2/(2 * sig**2)))
 
 # Sample 10000 points from a N(15.0, 1.5) distrubution
+np.random.seed(seed=123456789)
 sample = np.random.normal(loc=15.0, scale=1.5, size=(10000,))
 ydata, bin_edges = np.histogram(sample, 100)
 xdata = (bin_edges[1:] + bin_edges[:-1])/2


examples/gaussian.py
```diff
@@ -15,6 +15,7 @@
 model = A*exp(-((x - x0)**2/(2 * sig**2)))
 
 # Sample 10000 points from a N(15.0, 1.5) distrubution
+np.random.seed(seed=123456789)
```
pckroon: I get the feeling that specifying a seed for an example is a bit strange. I would prefer setting a better initial guess --- this way it seems you can only use symfit if you seed the random generator.

Jhsmit: At least it works now - I think the bigger issue is symfit's sensitivity to inital guesses. See also #120, if scipy can do the job and symfit not something is wrong.

tBuLi: I agree that it would make more sense to provide a better guess, although due to the Heisenbug nature of this both might be necessary.

tBuLi: @Jhsmit Do you know if providing a better guess would already solve the problem? If that alone is not enough I will merge this PR with the seed but I would prefer a fix without the seed.

Jhsmit: It does, but you have to guess 300, a guess value of 200 is not good enough. See also #120, where scipy's curve fit can fit with 200 as initial guess. Contrary to the finding in that issue, if I switch to ``ConstrainedNumericalLeastSqaures`` here, it doesnt solve the "issue", fitting still sometimes (depending on seed) fails with guess 200. I'll have another look at #120 if I have time later.

tBuLi: #120 has been solved on master, does that also solve the Heissenbug nature of this one to some extend? (I tried to form a new release, but that gave a new bug unfortunately so stay tuned.)

Jhsmit: It might have improved but its hard to tell because of the probabilistic nature of fitting. It sometimes works, sometimes not, roughy 1/3 times not.

PR #130, opened 2017-05-31
N-dimensional interactive guess

pckroon:
This allows for interactive guessing with more than one independent variable.
There's still some code in the `__init__` method that could be split out into the two strategies, but I'm not sure it's worth the effort. Also, the strategies and the InteractiveGuess object are currently very tightly interwoven; maybe the signature of the strategy method needs to be expanded so they get all the data they need, rather than accessing interactive guess's attributes.

Also, the ND update plot method is a bit too slow, it probably needs some numpy magic. Plus, more tests also won't hurt.


symfit/contrib/interactive_guess/interactive_guess.py
```diff
@@ -8,62 +8,86 @@
...
 #    raise EnvironmentError("Your symfit version is not supported. YMMV.")
 
 
-class InteractiveGuess2D(TakesData):
+class InteractiveGuess(TakesData):
     """A class that provides an graphical, interactive way of guessing initial
     fitting parameters."""
 
-    @keywordonly(n_points=100)
+    @keywordonly(n_points=50, log_contour=True, percentile=[5, 95])
```
tBuLi: [5, 95] should be a tuple or None. http://python-guide-pt-br.readthedocs.io/en/latest/writing/gotchas/?highlight=default%20

symfit/contrib/interactive_guess/interactive_guess.py
```diff
@@ -8,62 +8,86 @@
...
-        # Stretch the plot 10-20% in the X direction, since that is visually more
-        # appealing. We can't evaluate the model for x < x_initial, so don't.
+        # Stretch the plot 10-20% in the X direction, since that is visually
+        # more appealing. We can't evaluate the model for x < x_initial, so
+        # don't.
         for x in self.model.independent_vars:
             plotrange_x = x_maxs[x.name] - x_mins[x.name]
-            if not hasattr(self.model, 'initial'):
+            # if not hasattr(self.model, 'initial'):
+            if not isinstance(self.model, ODEModel):
```
tBuLi: why treat ODEModel seperate here? I'm not sure they have to be here, but a comment is missing.

symfit/contrib/interactive_guess/interactive_guess.py
```diff
@@ -206,4 +239,126 @@ def __str__(self):
...
+        plot_line.set_data(x_plot_data, y_plot_data)
+        error_lines[1].set_segments(y_segments)
+
+
+class InteractiveGuess2D(InteractiveGuess):
+    def __init__(self, *args, **kwargs):
+        import warnings
+        # Deprecated as of 01/06/2017
+        warnings.warn("InteractiveGuess2D is deprecated in favor of InteractiveGuess")
+        # raise DeprecationWarning("InteractiveGuess2D is deprecated in favor of InteractiveGuess")
```
tBuLi: Why not this DeprecationWarning so it is a proper subclass of DeprecationWarning rather than just a warning?

pckroon: Because raised DeprecationWarnings are not shown in normal operation, IIRC.

symfit/contrib/interactive_guess/interactive_guess.py
pckroon: Because you can't evaluate ODEModels at `t < t_initial`. A comment seems like a good idea.

symfit/contrib/interactive_guess/interactive_guess.py
pckroon: Correct. Good catch. I doesn't actually matter in this specific case, but still.

symfit/contrib/interactive_guess/interactive_guess.py
tBuLi: Good point, `ODEModel`'s should be extended to allow this. I opened #135 for this.

pckroon:
Rebased against master

pckroon:
Finally looks good from my end :)

symfit/contrib/interactive_guess/interactive_guess.py
```diff
@@ -8,53 +8,71 @@
...
 import matplotlib.pyplot as plt
 import numpy as np
+from scipy.stats import gaussian_kde
 
 plt.ioff()
 
 #from pkg_resources import parse_version
 
-#SUPPORTED_VERSION = '0.3.0'
+#SUPPORTED_VERSION = '0.3.6'
```
tBuLi: I'm not sure this is actually all that relevant, since interactive_guess is shipped with symfit anyway. So it should always support the version it is packaged with, and using it with a different version of symfit would require a lot of hacking, and if a user does that they deserve whatever happens to them.

symfit/contrib/interactive_guess/interactive_guess.py
```diff
@@ -206,4 +237,129 @@ def __str__(self):
         msg = 'Guessed values:\n'
         for param in self.model.params:
             msg += '{}: {}\n'.format(param.name, param.value)
-        return msg
\ No newline at end of file
+        return msg
+
+
+class Strategy2D:
```
tBuLi: What are strategies? Needs a docstring

symfit/contrib/interactive_guess/interactive_guess.py
tBuLi: Consider using the `@deprecated` decorator from this from `support.py`. It makes this process very easy.

symfit/contrib/interactive_guess/interactive_guess.py
```diff
@@ -206,4 +237,129 @@ def __str__(self):
...
+        xs = np.column_stack((x_plot_data, x_plot_data))
+        yerr = y_plot_error + y_plot_data[:, np.newaxis]
+        y_segments = np.dstack((xs, yerr))
+        plot_line, caps, error_lines = self.ig._plots[(indep_var, dep_var)]
+        plot_line.set_data(x_plot_data, y_plot_data)
+        error_lines[1].set_segments(y_segments)
+
+
+class InteractiveGuess2D(InteractiveGuess):
+    def __init__(self, *args, **kwargs):
```
tBuLi:
```suggestion
    @deprecated(InteractiveGuess)
    def __init__(self, *args, **kwargs):
```

symfit/contrib/interactive_guess/examples/simple_3D_example.py
```diff
@@ -0,0 +1,31 @@
...
+sig_x = Parameter('sig_x', 1)
+sig_y = Parameter('sig_y', 1)
+
+
+
+model = Model({z: Gaussian(x, mu_x, sig_x) * Gaussian(y, mu_y, sig_y)})
+x_data = np.linspace(0, 25, 50)
+y_data = np.linspace(0, 25, 50)
+x_data, y_data = np.meshgrid(x_data, y_data)
+x_data = x_data.flatten()
```
tBuLi: Why does the user have to flatten? I think that is a bit unnatural, personally as a user I'm very fearful of flatten so I would never think of typing this.

tBuLi:
Very nice improvements to the original features. What I'm still missing though is a description of the 3D fit; could you add an example similar to this one [here](https://symfit.readthedocs.io/en/stable/examples/ex_interactive_guesses_vector_2D.html) for your 3D example, which explains how the interface is to be used?

symfit/contrib/interactive_guess/interactive_guess.py
tBuLi: All these comments should still be removed.

symfit/contrib/interactive_guess/examples/simple_3D_example.py
tBuLi: Without the flatten it raises an error, but I don't remember from where. But I also don't agree with the analysis, if you plug in a 2d array for x and y, z will also be 2d for this model. And the reason I'm always fearful of using flatten is because it could lead to contracting the wrong shapes so long as the flattened array is the same size, whereas without flatten you will get a healthy broadcasting error if the shapes are not alligned.

pckroon: Hmmn, ok. I see your point. Still though, the model does `f(R, R) -> R`, not `f(R^2, R^2) -> R^2`. I'll try to hunt down where this goes wrong, and whether I can circumvent it.
What I want/need to do is for every independent variable, create `n` linearly spaced points, and evaluate the model on those.
At the same time, I need to make a contourplot of the given `z` data for every projection. This means that I need to do this:
```
projected_data = np.column_stack((x_data, z_data)).T
kde = gaussian_kde(projected_data)
```
and also for `(y_data, z_data)`. And `projected_data`has to be 1 or 2D: https://docs.scipy.org/doc/scipy-0.15.1/reference/generated/scipy.stats.gaussian_kde.html

So no, there's no way around flattening the data, and in that case I'd rather let the user do it.

pckroon: In addition, it's needed for plotting.

symfit/contrib/interactive_guess/interactive_guess.py
```diff
@@ -176,24 +206,17 @@ def _update_plot(self, _):
...
-        evaluated_model = self._eval_model()
         for indep_var, dep_var in self._projections:
-            plot = self._plots[(indep_var, dep_var)]
-            # TODO: reduce dimensionality of self._x_points and vals for this projection
-            y_vals = getattr(evaluated_model, dep_var.name)
-            x_vals = self._x_points[indep_var]
-            plot.set_data(x_vals, y_vals)
-#        self.fig.canvas.draw()  # Force redraw
+            self._update_specific_plot(indep_var, dep_var)
+        # self.fig.canvas.draw()  # Force redraw
```
tBuLi: unnecessary commented line?

symfit/contrib/interactive_guess/interactive_guess.py
```diff
@@ -86,18 +104,21 @@ def execute(self, **kwargs):
...
+        Any additional keyword arguments are passed to
+        matplotlib.pyplot.show().
         """
         show = kwargs.pop('show')
         if show:
             # self.fig.show()  # Apparently this does something else,
             # see https://github.com/matplotlib/matplotlib/issues/6138
             plt.show(**kwargs)
-            plt.close(self.fig)
+            #plt.close(self.fig)
```
tBuLi: Remove this comment?

PR #132, opened 2017-06-16
Negative fixed guess

### LINKED ISSUE #131: Setting a negative fixed parameter
I am needing to set a negative fixed parameter for a model. This may be an issue or user error. I started the following stack exchange question. 

https://stackoverflow.com/questions/44562190/setting-a-negative-fixed-parameter-in-the-symfit-python-package

Sorry for leaving a comment on Git, but both Stack Exchange and Git removed the ability to send comments so I figured this would be the best way to alert those who work on the project as Stack Exchange does not have a Symfit tag.

tBuLi:
Fixes #131 for future generations

### DIFF
--- a/symfit/core/argument.py
+++ b/symfit/core/argument.py
@@ -88,8 +88,13 @@ class Parameter(Argument):
         self.value = value
         self.fixed = fixed
         if not self.fixed:
-            self.min = min
-            self.max = max
+            if min is not None and max is not None and min > max:
+                print(min, max)
+                raise ValueError(
+                    'The value of `min` should be less than or equal to the value of `max`.')
+            else:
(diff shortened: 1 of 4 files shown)


tex-downey:
tBuLi,

This equation (with the negative parameters) also through some odd errors when I tried to set mins and maxs for my parameters.  I see you updated somthing related to that just wanted to make sure you were aware of this issue.

tBuLi:
Are those warnings or errors? I am aware that you can get some warnings, but as long as the end result is a good fit those are nothing to worry about.

tex-downey:
They were errors. But changing Fit to NumericalLeastSquares seemed to have resolved the issue.

Issue #56, opened 2016-05-03
DOI number for symfit


pckroon:
It should be possible to easily cite symfit in academic material, so it should have an DOI number. Luckily, this shouldn't be to hard according to https://guides.github.com/activities/citable-code/.

tex-downey:
I agree, I have cited the documentation page but a DOI would be nice. This should be a simple process for one of the main contributors.

tBuLi:
You are right, I have started setting everything up. In principle the next release should trigger the creation of a DOI, so stay tuned!

pckroon:
Is there an ETA for 0.4.0?

Issue #106, opened 2016-12-15
Max iteration keyword differs for different fit objects


tBuLi:
When calling the new smart Fit object, the name of the maximum iteration keyword passed to scipy has to be different for NumericalLeastSquares and ConstrainedNumericalLeastSquares. That shouldn't be the case of course.

All common keywords should therefore be wrapped. Since this package aims for readability, I propose to use clear and descriptive names such as `max_iterations` rather than copy the obscure scipy ones.

PR #133, opened 2017-06-29
Refactor Fit objects

tBuLi:
This  PR represents a mayor overhaul of the internal structure of Fit objects in symfit.

`Fit` is now responsible for feeding an Objective function to a Minimizer, it is no longer the combination of these two.

This means two new types of object have been introduced; Minimizers and Objectives. This change should make it easier for people to introduce a new Minimizer while being completely agnostic about which objective function to use.

To do (not a complete list, work in progress):
- Documentation overhaul
- Developers instructions in the docs
- Tests
- Make keyword arguments consistent across minimizers (e.g. #106 )
- Covariance matrix for all types of models including Log-likelihood (#138)
- Wrap all minimizers in scipy.optimize.minimize with the same new API.
- ...


pckroon:
We need to address one more thing in this fell swoop, and that is to decide how the arguments are called that fit.execute (and thereby Minimizer.execute) is to accept, and what it should return. In the process we should standardise some names.

tBuLi:
The error that now remains in this branch is one that will be solved my merging with @pckroon's finite differences branch, because then stdev's for ODEModels can be calculated.

symfit/core/objectives.py
```diff
@@ -0,0 +1,216 @@
...
+                if flatten_components: # Flattens *within* a component
+                    result[-1] = result[-1].flatten()
+        return np.sqrt(sum(result))
+
+    def eval_jacobian(self, **parameters):
+        # jac_args = list(independent_data.values()) + list(p)
+        evaluated_func = self.model(**self.independent_data, **parameters)
+        result = [0.0 for _ in self.model.params]
+
+    def eval_jacobian(self, **parameters):
```
pckroon: Dubble function definition

symfit/core/objectives.py
```diff
@@ -0,0 +1,216 @@
...
+                    result[index] += np.sum(
+                        component(**self.independent_data, **parameters) * ((dep_data - ans) / sigma ** 2)
+                    )
+        return - np.array(result).T
+
+
+class LogLikelihood(GradientObjective):
+    def __call__(self, **parameters):
+        """
+        Error function to be maximised(!) in the case of log-likelihood fitting.
```
pckroon: Doesn't it make more sense to agree that all objectives should *always* be minimized, and flip the sign here?

tests/test_analytical_fit.py
```diff
@@ -187,13 +188,13 @@ def test_weights(self):
         fit_result = fit.execute()
 
         self.assertAlmostEqual(num_result.value(sqrt_g_inv), fit_result.value(sqrt_g_inv))
-        self.assertAlmostEqual(num_result.value(b), fit_result.value(b))
+        self.assertAlmostEqual(num_result.value(b) / fit_result.value(b), 1.0, 5)
```
pckroon: Elegant. I like it. Maybe set the precision (5) as a global variable for all the tests?

tests/test_analytical_fit.py
```diff
@@ -216,28 +217,28 @@ def test_backwards_compatibility(self):
...
+        fit = Fit(model, y=yn, sigma_y=sigma, absolute_sigma=False, minimizer=MINPACK)
         num_result = fit.execute()
 
         popt, pcov = curve_fit(lambda x, a: a * np.ones_like(x), xn, yn, sigma=sigma, absolute_sigma=False)
 
 
-        self.assertAlmostEqual(fit_result.value(a), num_result.value(a))
-        self.assertAlmostEqual(fit_result.stdev(a), num_result.stdev(a))
+        self.assertAlmostEqual(fit_result.value(a), num_result.value(a), 5)
+        self.assertAlmostEqual(fit_result.stdev(a), num_result.stdev(a), 5)
```
pckroon: Relative precision? i.e. divide?

tests/test_analytical_fit.py
```diff
@@ -216,28 +217,28 @@ def test_backwards_compatibility(self):
...
-        fit = NumericalLeastSquares(model, y=yn, sigma_y=sigma, absolute_sigma=True)
+        fit = Fit(model, y=yn, sigma_y=sigma, absolute_sigma=True, minimizer=MINPACK)
         num_result = fit.execute()
 
         popt, pcov = curve_fit(lambda x, a: a * np.ones_like(x), xn, yn, sigma=sigma, absolute_sigma=True)
 
-        self.assertAlmostEqual(fit_result.value(a), num_result.value(a))
-        self.assertAlmostEqual(fit_result.stdev(a), num_result.stdev(a))
+        self.assertAlmostEqual(fit_result.value(a), num_result.value(a), 5)
+        self.assertAlmostEqual(fit_result.stdev(a), num_result.stdev(a), 5)
```
pckroon: /

tests/test_fit_result.py
```diff
@@ -155,28 +114,21 @@ def test_fitting_2(self):
...
-            for warning in w:
-                self.assertTrue(issubclass(warning.category, DeprecationWarning))
 
+    def test_minimizer_included(self):
+        """"The minimizer used should be included in the results."""
+        return NotImplementedError()
+
+    def test_objective_included(self):
+        """"The objective used should be included in the results."""
+        return NotImplementedError()
```
pckroon: Excellent ideas. That way you could even conceive letting the minimizer (or objective?) keep track of the number of evaluations. Let's leave it for later though.

tests/test_general.py
```diff
@@ -78,17 +81,18 @@ def test_vector_fitting(self):
...
             c_i=xdata[2],
+            minimizer = MINPACK
         )
         fit_result = fit.execute()
 
-        self.assertAlmostEqual(fit_result.value(a), 9.985691, 6)
-        self.assertAlmostEqual(fit_result.value(b), 1.006143e+02, 4)
-        self.assertAlmostEqual(fit_result.value(c), 7.085713e+01, 5)
+        self.assertAlmostEqual(fit_result.value(a) / 9.985691, 1.0, 5)
+        self.assertAlmostEqual(fit_result.value(b) / 1.006143e+02, 1.0, 4)
```
pckroon: 5?

tests/test_general.py
```diff
@@ -752,7 +759,8 @@ def test_error_analytical(self):
 
         fit = Fit(model, y=yn, sigma_y=sigma)
         fit_result = fit.execute()
-
+        print(fit_result)
+        print(fit.absolute_sigma)
```
pckroon: leftover debug output?

tests/test_minimizers.py
```diff
@@ -0,0 +1,83 @@
...
+        fit_result = fit.execute()
+
+        def f(x, a, b):
+            return a * x + b
+
+        def chi_squared(a, b):
+            return np.sum((ydata - f(xdata, a, b))**2)
+
+        fit_custom = BFGS(chi_squared, [a, b])
+        fit_custom_result = fit_custom.execute()
```
pckroon: `fit_custom = Fit(model, xdata, ydata, minimizer=BFGS(chi_squared, [a, b]))`? I probably have the signature wrong here though.

tests/test_new_fit.py
```diff
@@ -0,0 +1,36 @@
...
+fit_result = fit.execute()
+print(fit_result)
+
+y = model(x=xdata, **fit_result.params)
+# # sns.regplot(xdata, ydata, fit_reg=False)
+# plt.plot(xdata, y)
+# plt.scatter(xdata, ydata, c='r')
+# # plt.xlim(0, 100)
+# # plt.ylim(0, 2000)
+# plt.show()
```
pckroon: This file looks like a debug test, and should probably be removed?

pckroon:
Looking good!
Note that you should probably fix #56 before tagging and releasing.

symfit/api.py
```diff
@@ -3,10 +3,11 @@
...
-    Fit, FitResults, Maximize, Minimize, Likelihood, Model,
-    NumericalLeastSquares, LinearLeastSquares, NonLinearLeastSquares,
-    TaylorModel, ODEModel, ConstrainedNumericalLeastSquares, ModelError
+    Fit, Model, Constraint,
+    LinearLeastSquares, NonLinearLeastSquares,
+    TaylorModel, ODEModel, ModelError
 )
+from symfit.core.fit_results import FitResults
 from symfit.core.argument import Variable, Parameter
 from symfit.core.support import variables, parameters, D
```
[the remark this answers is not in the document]
tBuLi: From looking at [this](https://docs.python.org/2.5/whatsnew/pep-328.html) it seems relative imports are supported since 2.5, so that is possible

symfit/core/fit_results.py
```diff
@@ -0,0 +1,118 @@
...
+    def __getattr__(self, item):
+        """
+        Return the requested `item` if it can be found in the gof_qualifiers
+        dict.
+
+        :param item: Name of Goodness of Fit qualifier.
+        :return: Goodness of Fit qualifier if present.
+        """
+        if item in self.gof_qualifiers:
+            return self.gof_qualifiers[item]
```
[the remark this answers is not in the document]
tBuLi: Actually yes, that is needed. Now it is essentially ` else: return None`.

symfit/core/fit.py
```diff
@@ -1160,14 +951,15 @@ def execute(self):
...
+            model=self.model,
             popt=[best_fit_params[param] for param in self.model.params],
             pcov=cov_matrix,
             infodic={'nfev': 0},
             mesg='',
             ier=0,
         )
-        self._fit_results.r_squared = r_squared(self.model, self._fit_results, self.data)
+        self._fit_results.gof_qualifiers['r_squared'] = \
+            r_squared(self.model, self._fit_results, self.data)
```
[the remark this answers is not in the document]
tBuLi: Actually, these objects should probably also be recast into the new paradigm, although some thinking is needed on how to do this. Basically, these are analytically exact least-square solvers, not general purpose minimizers. But perhaps all this functionality should in fact be stripped from these objects and taken over by Fit, just like we do for other Minimizers.

symfit/core/fit.py
```diff
@@ -1235,93 +1027,147 @@ def execute(self, relative_error=1e-8, max_iter=500):
...
-        :constraints: constraints the minimization is subject to.
-        :type constraints: list
+        :param model: (dict of) sympy expression(s) or ``Model`` object.
+        :param constraints: iterable of ``Relation`` objects to be used as
+            constraints.
+        :param bool absolute_sigma: True by default. If the sigma is only used
+            for relative weights in your problem, you could consider setting it to
+            False, but if your sigma are measurement errors, keep it at True.
+            Note that curve_fit has this set to False by default, which is wrong in
+            experimental science.
```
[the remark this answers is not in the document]
tBuLi: `absolute_sigma` is kicked up to `TakesData`, who handles it. The keyword_only decorator is such that if you inspect this `__init__` it will actually contain `absolute_sigma` as well, and so it is sphynx friendly ;).

symfit/core/fit.py
```diff
@@ -1235,93 +1027,147 @@ def execute(self, relative_error=1e-8, max_iter=500):
...
+        # List of Constraint objects
+        self.constraints = self._init_constraints(constraints=constraints)
+
+        if objective is None:
+            # Param only scalar Model -> the model is the objective.
+            if len(self.model.independent_vars) == 0 and len(self.model) == 1:
+                # No data provided means a simple minimization of the Model parameters
+                # is requested, not a fit.
+                if all(value is None for value in self.data.values()):
+                    objective = MinimizeModel
```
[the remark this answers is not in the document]
tBuLi: I agree that currently it could be `and`-ed , but since both are already very long statements I don't think it will help readability.

An nice example of something that doesn't match the second if btw would be `y = a`, which means that `a` will be the mean value of `y`.  Nice eh?

symfit/core/fit.py
```diff
@@ -1235,93 +1027,147 @@ def execute(self, relative_error=1e-8, max_iter=500):
...
+        # Initialise the minimizer
+        if isinstance(minimizer, BaseMinimizer):
+            self.minimizer = minimizer
+        else:
+            minimizer_options = {}
+            if issubclass(minimizer, GradientMinimizer):
+                # if isinstance(self.model, ODEModel):
+                #     raise Exception(self.model.numerical_jacobian)
+                if hasattr(self.model, 'numerical_jacobian') and hasattr(self.objective, 'eval_jacobian'):
+                    minimizer_options['jacobian'] = self.objective.eval_jacobian
```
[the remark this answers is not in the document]
tBuLi: Yes, because if no jacobian is provided explicitly it will be estimated numerically. So only if the model provides an analytical jacobian should `eval_jacobian` be preferred over letting the minimizer work it's magic. Actually, this goes back nicely to our discussion that `eval_jacobian` calls with finite differences are relatively expencive. I think that is why you choose to check for both `eval_jacobian` and `numerical_jacobian`, the latter indicating if it has an analytical form.

I will add all this as a comment to the code though.

symfit/core/fit.py
```diff
@@ -1235,93 +1027,147 @@ def execute(self, relative_error=1e-8, max_iter=500):
...
+
+            if issubclass(minimizer, ConstrainedMinimizer):
+                if issubclass(minimizer, ScipyMinimize):
+                    minimizer_options['constraints'] = minimizer.scipy_constraints(
+                        self.constraints,
+                        self.data
+                    )
+                else:
+                    minimizer_options['constraints'] = self.constraints
```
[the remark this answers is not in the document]
tBuLi: Good point, but we decided that Minimizers should be agnostic about data. They simply minimize the objective function by playing with the values of the parameters, that's it. So this initiation can not be done by the Minimizer unfortunately. 

Arguably the data doesn't have to be there though, I don't think any of the current test scenarios has constraints which are data dependent. But I recently needed this in my research actually so I think it is best to leave it like this.

symfit/core/fit.py
```diff
@@ -1332,547 +1178,35 @@ def __init__(self, model, *args, **kwargs):
...
-        """
-        Returns :math:`\\chi^2`, summing over all the vector components and
-        data indices.
+            if cov_matrix is None:
+                cov_matrix = self.covariance_matrix(dict(zip(self.model.params, minimizer_ans._popt)))
+        finally:
+            minimizer_ans.covariance_matrix = cov_matrix
+        # Overwrite the DummyModel with the current model
+        minimizer_ans.model = self.model
+        minimizer_ans.gof_qualifiers['r_squared'] = r_squared(self.model, minimizer_ans, self.data)
```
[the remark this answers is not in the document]
tBuLi: I'm happy to inform you that it is already there, though not visible from this code. The execute methods of the minimizers now always include the value of the objective at the minimum in the result. So if the objective function used was a chi_squared, it's included under the name `objective_value` in the `FitResults`.

symfit/core/minimizers.py
```diff
@@ -0,0 +1,217 @@
...
+class BaseMinimizer:
+    """
+    ABC for all Minimizers.
+    """
+    def __init__(self, objective, parameters, absolute_sigma=True):
+        """
+        :param objective: Objective function to be used.
+        :param parameters: List of :class:`~symfit.core.argument.Parameter` instances
+        :param absolute_sigma: Wether or not the standard deviations should be
+            interpreted as relative or absolute weights.
```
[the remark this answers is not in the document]
tBuLi: Very good point, neither. Covariances are calculated separately by `Fit`, such that it is consistent across methods.

symfit/core/objectives.py
tBuLi: Not sure I understand this comment, because the sign is indeed flipped by default here.
`ans = - np.nansum(np.log(self.model(**jac_kwargs)))`

This is more intuitive I think in this case. Or is your mark about the (exclamation point in the) docstring?

tests/test_analytical_fit.py
tBuLi: Yes, relative testing everywhere is probably the way to go.

tests/test_minimizers.py
tBuLi: Not sure what that would mean since it's a mix of both paradigms. Either we use chi_squared here to manually fit with BFGS, or we let Fit take care of it. or would you like to see the objective be made explicit when calling Fit? e.g. `objective=LeastSquares`

tests/test_new_fit.py
tBuLi: indeed

symfit/core/fit.py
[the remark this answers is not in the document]
pckroon: I think Fit should pick an objective, and then decide on the best way to minimize it. Analytical might be the answer. So in that sense, they *are* minimizer, just really funny ones. I think it should be possible to cast the in the new API.

symfit/core/fit.py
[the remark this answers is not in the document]
pckroon: Should Minimizers then have a static method that takes symfit constraints and translates them to a form it recognizes? We can leave this as is for now, and leave a big fat comment.

symfit/core/objectives.py
pckroon: In that case the docstring lies, because it says it should be maximised.

tests/test_minimizers.py
pckroon: I was thinking to initialise Fit with an instantiates minimizer. But I forgot what the point was.
In this case I do think it's also important to specify the objective.

PR #136, opened 2017-07-12
Finite difference

pckroon:
Implements the finite difference method to approximate the jacobian when no analytical solution is available, and use that for e.g. `ODEModels` to get estimates for the standard deviation in the parameters.

This also moved some methods from `Model` to `CallableModel`, where I thought it would make sense, and added some logic to `Model`'s `eval_jacobian` to make sure it get a sane shape.

### DIFF
--- a/symfit/api.py
+++ b/symfit/api.py
@@ -3,10 +3,11 @@ import symfit.core.operators
 
 # Expose useful objects.
 from symfit.core.fit import (
-    Fit, FitResults, Maximize, Minimize, Likelihood, Model,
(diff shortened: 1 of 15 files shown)


pckroon:
I fixed most test failures, but there's one still there (unequal data length). I'm not sure what to do with that one. In addition, I put a skeleton in the closet of `HasCovarianceMatrix._cov_mat_unequal_lengths`.
Both need some attention and love before this is merged.

pckroon:
I made a new PR which aim to merge these changes into your refactor branch (#137). Let's continue there.

PR #137, opened 2017-11-28
Finite difference

pckroon:
There's still two failing tests where the stdev in ODEModels is nan, even though the test checking that `np.isfinite(ode_result.stdev(...))` succeeds. I need anther pair of eyes.


tBuLi:
So I have made some interesting observations.

By making some changes on how to deal with the dependent vars whose data is set to `None`, the covariance matrix now seems to be calculated correctly for those `ODEModel`'s.

However, for the simplest model of a simple exponential model, the covariance matrix is off by a factor of 2, and thus the stdev by a factor of sqrt(2). This is a big mystery to me, since the tests on the finite_difference method indicate that this method is in very good agreement with the cases where the jacobian is calculated exactly. So that strongly suggests the problem isn't there.

What's most frustrating is that the test for the two component model does seem to yield the correct answer, though this answer is hardcoded so I don't know where it came from.

So the next steps should to find more exactly solvable toy-models with multiple components, and to test those instead.

Issue #60, opened 2016-07-22
you have to give name kwargs to parameters/variables in jupyter notebooks


Jhsmit:
Havent tested ipython but in jupyter you cant make paramters without using the name kwargs. This makes sense but it could be nice to give a useful error message to the user urging him to specify 'name'.

Pitje06:
IIRC this happens in all interactive sessions, and is due to problems with the introspection module not working properly there. 
I don't know if we can catch this specific error, or should submit it somewhere upstream.

Jhsmit:
It seems to be possible to detect if python is in interactive mode:

```python
def is_interactive():
    import __main__ as main
    return not hasattr(main, '__file__')
```
http://stackoverflow.com/questions/15411967/how-can-i-check-if-code-is-executed-in-the-ipython-notebook

Pitje06:
I can't reproduce the original issue in either Ipython 5.1.0, python 3.5.2, nor in a jupyter notebook 4.4.2. Can you still reproduce it?

Jhsmit:
Nope seems to work for me in jupyter notebook. Somebody must have fixed it.

tBuLi:
If there is one design choice I regret it's using inspect to guess the names. It introduces more problems than it solves.

I am definitely considering removing it, but still having the name be optional for backwards compatibility, and throwing a deprecation warning. If no name is provided, make it a sympy.dummy parameter, just like I already do with dependent variables if no name is provided.

Jhsmit:
Black magic is not to be used lightly. 

If a user doesnt bother with explicitly giving a name its not likely he will explicitly call it. However I'm wondering what a `print(result)` will look like when no name is known. Numbered parameters by order of creation?

pckroon:
Closed by #140

PR #140, opened 2017-12-29
Remove inspect, more sympy compatible style.

tBuLi:
Removed the inspect module usage to enforce DRY. This means the code is now more backwards compatible with sympy, at the expense of forcing users to write a little bit more code. However, the usage of inspect caused more problems then it solved. This might cure #60.

Now, if no name is provided, a dummy name is generated in the background, e.g.
```python
>>> a = Parameter()
>>> print(a)
p_0
```
```python
>>> a = Parameter('a')
>>> print(a)
a
```
```python
>>> a = Parameter('a', 5)
>>> print(a, a.value)
a, 5
```
```python
>>> a = Parameter(value=5)
>>> print(a, a.value)
p_0, 5
```
```python
>>> a = Parameter(5)
TypeError: In symfit >0.4.1 the value needs to be assigned as the second argument or by keyword argument.
```

I am now debating whether this is the way to go, or to just remove this entirely and to make users type out the names explicitly, just like `sympy`, since it prevents a lot of ambiguity.

### DIFF
--- a/symfit/core/argument.py
+++ b/symfit/core/argument.py
@@ -24,11 +24,12 @@ class Argument(Symbol):
         # Generate a dummy name
         if not name:
             # Throw a warning that is is better to explicitly give names.
-            warnings.warn(DeprecationWarning(
+            warnings.warn(
                 'It is recommended to provide names to {} explicitly'
                 ' as automatic generation of names will be dropped in '
-                'future `symfit` versions.'.format(cls.__name__)
-            ))
+                'future `symfit` versions.'.format(cls.__name__),
+                DeprecationWarning, stacklevel=2
+            )
 
             name = '{}_{}'.format(cls._argument_name, cls._argument_index)
             instance = super(Argument, cls).__new__(cls, name, **assumptions)
@@ -49,7 +50,7 @@ class Argument(Symbol):
 class Parameter(Argument):
     """
     Parameter objects are used to facilitate bounds on function parameters.
-    Important change from `symfit>=0.4.1`: the name needs to be the first keyword,
+    Important change from `symfit>0.4.1`: the name needs to be the first keyword,
     followed by the guess value. If no name is provided, the initial value can
     be passed as a keyword argument, e.g.: `value=0.1`. A generic name will then
     be generated.
--- a/symfit/core/argument.py
+++ b/symfit/core/argument.py
@@ -1,4 +1,5 @@
 import numbers
+import warnings
 
 from sympy.core.symbol import Symbol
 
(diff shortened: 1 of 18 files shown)


symfit/core/argument.py
```diff
@@ -1,70 +1,36 @@
...
-        return super(Argument,cls).__new__(cls, name, **assumptions)
+        # Generate a dummy name
+        if not name:
+            generated_name = '{}_{}'.format(cls._argument_name, cls._argument_index)
+            instance = super(Argument, cls).__new__(cls, generated_name, **assumptions)
+            instance._argument_index = cls._argument_index
+            cls._argument_index += 1
+            return instance
+        else:
+            return super(Argument, cls).__new__(cls, name, **assumptions)
```
pckroon: Isn't it better to let `super().__new__` (i.e. sympy) generate a name? Otherwise:
```
if not name:
    name = '{}_{}'.format(cls._argument_name, cls._argument_index)
    cls._argument_index += 1
instance = super(Argument, cls).__new__(cls, name, **assumptions)
instance._argument_index = cls._argument_index  # Do we need to do this? I think not.

```

symfit/core/support.py
```diff
@@ -87,23 +87,27 @@ def f(x, p):
...
     Convenience function for the creation of multiple parameters.
 
-    :param names: string of parameter names. Should be comma seperated.
+    :param names: string of parameter names.
         Example: a, b = parameters('a, b')
+    :param kwargs: kwargs to be passed onto :meth:`sympy.symbols`
+    :return: iterable of :class:`symfit.core.Parameter` objects
     """
-    return [Parameter(name=name.strip()) for name in names.split(',')]
+    return symbols(names, cls=Parameter, seq=True, **kwargs)
```
pckroon: You can do some meta-programming thing here by making a symbol_builder which you pass just `Parameter` or `Variable`. It would save you a little repetition, but probably won't be worth the effort.

tests/test_argument.py
```diff
@@ -0,0 +1,87 @@
...
+    parameters, Model, Eq, Ge
+)
+from symfit.core.minimizers import BFGS, MINPACK, SLSQP, LBFGSB
+from symfit.core.objectives import LogLikelihood
+from symfit.distributions import Gaussian, Exp
+
+if sys.version_info >= (3, 0):
+    import inspect as inspect_sig
+else:
+    import funcsigs as inspect_sig
```
pckroon: Imports need to be trimmed

tBuLi:
I think a number as the first argument has to raise an error because otherwise it becomes ambiguous what the first argument to the function is: a name or the value. That should not be type dependent.

I do still think that maybe it's better to rip this out entirely and go back to sympy style. Or alternatively, do it as I have in this PR but update all the docs to give parameters names explicitly. That way we keep backwards compatibility but educate people on the preferred style, which we will then switch to in the future. I could even include a deprication warning when no name is provided.

pckroon:
What's "sympy style"? Updating the docs is a good idea. You can do that already ;)
Deprecation warnings are also always a good idea.

symfit/core/support.py
tBuLi: I'm actually thinking of going back to the `sympy` essence; which is to remove `parameters` and `variables` entirely and to update the docs to have users call `symbols(names, cls=Parameter)` directly. It's only slightly more verbose for the user and does provide a lot of clarity.

symfit/core/support.py
```diff
@@ -87,9 +112,11 @@ def f(x, p):
 
     return f
 
+@deprecated('symbols(names, cls=Variable)')
 def variables(names, **kwargs):
```
pckroon: I disagree with the deprecation of variables and parameters. These are convenient helper functions

symfit/core/support.py
```diff
@@ -98,9 +125,11 @@ def variables(names, **kwargs):
     """
     return symbols(names, cls=Variable, seq=True, **kwargs)
 
+@deprecated('symbols(names, cls=Parameter)')
 def parameters(names, **kwargs):
```
pckroon: Same.

Issue #144, opened 2018-03-24
Singular Covariance Matrix crash


tBuLi:
If the covariance matrix happens to be a singular matrix, numpy throws `LinAlgError: Singular matrix`, which in turn crashes symfit.

I do not currently have time to write a minimal example of how to reproduce this, but it shouldn't take more then a try/except to fix since this is a thing that can just happen depending on the data.

A more nuanced fix though, would involve checking if the minimizer used returns a hessian_inv, and use that instead of computing it anew as we currently do in symfit 0.4.0. The reason for computing it anew is to stay consistent across minimizers, as well as to not have to deal with the inconsistent API of the scipy  solvers. But maybe it is worth doing that for a maximum chance of success.

pckroon:
Is the hessian_inv exactly identical to the covariance matrix? Or is there still some scaling of sorts involved? If the former, I'll have a look.

tBuLi:
There is a scaling involved, depending on whether the weights are to be interpreted as measurements errors or not. 

If we interpret as measurement errors then we need to rescale with the total sum of residuals as also remarked here: https://stats.stackexchange.com/questions/71154/when-an-analytical-jacobian-is-available-is-it-better-to-approximate-the-hessia (Sorry can not find a better source right now)

But this can be achieved out of the box by copying [this](https://github.com/tBuLi/symfit/blob/4ef619491a9979e1e3371047b23ba9739fa4207d/symfit/core/fit.py#L908) if/else statement.

tBuLi:
p.s. it also worth rereading my final answer in #138 since it seems there was an issue with this. Maybe it is not as trivial as it seems, or I forgot precisely this rescaling when testing for #138. I don't have time to go through all the sources right now though.

pckroon:
That SO post is good :)
It might be worthwhile to separate the case where we have an analytical Jacobian from the one where we don't.
To summarize:

1) We have the analytical Jacobian
In this case it's best to either derive the Hessian analytically, or calculate it via finite differences.

2) We do not have the analytical Jacobian
Here we should use the (inversed) Hessian given by the minimizer (if we have it), and otherwise it's probably best to compute J^T*J and pray the residuals are small enough. If that's not good enough (for you), it should be possible to calculate a Hessian via finite differences on the original function. That might be expensive though.

tBuLi:
[...]

PR #145, opened 2018-04-01
Small fixes due to changes upstream. Fixed issue with fixed parameters. Quickfix for #144

pckroon:
Apparently there are some changes upstream in Scipy.minimize and matplotlib. In addition, fixed parameters did not get min and max attributes, while they are required by the current minimizers. Lastly, I made the quick fix for #144. I'm going to work on fixing it proper about now.


pckroon:
I messed up with Git, polluting this with new features. New one incomig.

PR #146, opened 2018-04-01
Fix upstream compatibility

pckroon:
In addition, a quick fix for #144, and made sure that all parameters will get min and max attributes, even if they're fixed.


symfit/core/minimizers.py
```diff
@@ -102,7 +102,7 @@ def execute(self, bounds=None, jacobian=None, **minimize_options):
             bounds=bounds,
             constraints=self.constraints,
             jac=self.wrapped_jacobian,
-            **minimize_options
+            options=minimize_options
```
tBuLi: I disagree with this line, because there might be more kwargs then just options.

tBuLi:
So if I understand correctly, scipy changed the name of the maxiter argument? However, just changing it in place like this might introduce problems for everyone on an older version of scipy right? Perhaps this should be solved in accordance with #106?

symfit/core/minimizers.py
tBuLi: The best alternative is the way it was :). If they ever add a kwarg to minimize, we will have to have this discussion again. I agree that having to explicitly say `options={maxiter: 1000}` in your calls sucks, but that is what #106 should solve someday.

symfit/core/fit.py
```diff
@@ -845,11 +845,14 @@ def covariance_matrix(self, best_fit_params):
...
-        else:
-            return self._cov_mat_unequal_lenghts(best_fit_params=best_fit_params)
+        try:
+            if len(set(arr.shape for arr in self.sigma_data.values())) == 1:
+                # Shapes of all sigma data identical
+                return self._cov_mat_equal_lenghts(best_fit_params=best_fit_params)
+            else:
+                return self._cov_mat_unequal_lenghts(best_fit_params=best_fit_params)
+        except np.linalg.LinAlgError:
+            return None
```
tBuLi: nice

PR #147, opened 2018-04-01
First implementation of Hessians by finite differences

pckroon:
Includes commits from #146.
This is a first implementation of hessian matrices by both finite difference methods, and analytical. The code still needs quite a bit of love, and there are issues with rounding/truncation errors.
And to put it out there, how do you feel about Complex Step Differentiation (CSD, https://blogs.mathworks.com/cleve/2013/10/14/complex-step-differentiation/)?

If you have any good idea on how to make this code readable and understandable, please let me know. If you don't and this is good enough for you, please merge it.
Either way, you can start building Covariance matrices from this.


tBuLi:
I think it would be a easier to implement this using 
hessian
because all the tools on the right hand side of this equation already exist. [source](https://calculus.subwiki.org/wiki/Hessian_matrix#Definition_in_terms_of_Jacobian_matrix_and_gradient_vector)

I'm just not sure how this changes for vector valued functions, but I think it will still be the same. It will just mean that the hessian is not a 2-rank tensor (matrix) but 3-rank, but that won't change a thing. Just use np.tensordot instead of dot. This will make the implementation a lot cleaner as you can just call model.eval_jacobian (?), which already decides for itself it wants to be analytic or finite-difference.

I absolutely love the CSD btw, that is a beautiful thing! This is why people need to stop being scared of complex numbers. Complex numbers are your friend! Definitely worth looking into, the best test case that comes to mind is ODEModel, since for those models we can in principle evaluate F(x; a + ih), but cannot directly give the value of dF/da, where a is a fit parameter. Having said that, I don't think the current ODE solver likes complex numbers, but mathematically I don't see a fundamental problem with it.

pckroon:
This simply does not do what we need/want, so closing.

PR #151, opened 2018-05-01
Make minimizers deal with fixed parameters

### LINKED ISSUE #150: Clarification on Fixing the Value of Parameter
For the following example, I was hoping to get some clarification on the parameter c.  My reading of this syntax is that the value of c should not change through the fitting routine.  However, upon examination of the fit results it appears as though it has, in fact, be altered.  Is there a way to truly fix the value so it doesn't change or am I simply misunderstanding the approach? Currently, I'm using symfit 0.4.0 and python 3.

```
from symfit import parameters, variables, exp, Fit, Parameter

a, b, c, d = parameters('a, b, c, d')
x, y = variables('x, y')

c.value = 4.0
c.fixed = True

model_dict = {y: a * exp(-(x - b)**2 / (2 * c**2)) + d}
fit = Fit(model_dict, x=xdata, y=ydata)
fit_result = fit.execute()
```
The output of the fit_result.params call is and I would expect c to still be 4.0:

```
OrderedDict([('a', 16.374368575343127),
             ('b', 0.49201249437123556),
             ('c', 0.5337962977235504),
             ('d', -9.55593614465743)])
```

Cheers,
Brian

pckroon:
Fixes #150.

The first part of this PR was accidentally pushed to master directly... Git is hard.
I applied my git wizard, and all should be well now.

### DIFF
--- a/symfit/core/minimizers.py
+++ b/symfit/core/minimizers.py
@@ -67,22 +67,31 @@ class GradientMinimizer(BaseMinimizer):
     def __init__(self, *args, **kwargs):
         jacobian = kwargs.pop('jacobian')
         super(GradientMinimizer, self).__init__(*args, **kwargs)
-        def _jacobian(*args, **kwargs):
-            # Assume the order of what jacobian returns and parameters is the same
-            # We need to remove the values for the fixed parameters from what
-            # jacobian returns
-            fixed_vals = {p.name: p.value for p in self._fixed_params}
-            kwargs.update(fixed_vals)
-            out = jacobian(*args, **kwargs)
+        jac_with_fixed_params = partial(jacobian, **{p.name: p.value for p in self._fixed_params})
+
+        if jacobian is not None:
+            self.wrapped_jacobian = self.wrap_jac(jac_with_fixed_params)
+        else:
+            self.jacobian = None
+            self.wrapped_jacobian = None
+
+    def wrap_jac(self, func):
+        """
+        Removes values with identical indices to fixed parameters from the
+        output of func.
+        :param func: Function to be wrapped
+        :return: wrapped function
+        """
+        if func is None:
+            return None
+        def wrapped(*args, **kwargs):
+            out = func(*args, **kwargs)
             jac = []
             for param, val in zip(self.parameters, out):
                 if not param.fixed:
                     jac.append(val)
             return jac
-        if jacobian is not None:
-            self.jacobian = _jacobian
-        else:
-            self.jacobian = None
+        return wrapped
 
 class ScipyMinimize(object):
     """
@@ -118,7 +127,7 @@ class ScipyMinimize(object):
             self.initial_guesses,
             bounds=bounds,
             constraints=self.constraints,
-            jac=self.wrapped_jacobian,
+            jac=jacobian,
             **minimize_options
         )
         # Build infodic
@@ -181,7 +190,7 @@ class ScipyMinimize(object):
 class ScipyGradientMinimize(ScipyMinimize, GradientMinimizer):
     def __init__(self, *args, **kwargs):
         super(ScipyGradientMinimize, self).__init__(*args, **kwargs)
-        self.wrapped_jacobian = self.wrap_func(self.jacobian)
+        self.wrapped_jacobian = self.wrap_func(self.wrapped_jacobian)
 
     def execute(self, **minimize_options):
         return super(ScipyGradientMinimize, self).execute(jacobian=self.wrapped_jacobian, **minimize_options)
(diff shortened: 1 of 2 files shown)


symfit/core/minimizers.py
```diff
@@ -20,8 +21,10 @@ def __init__(self, objective, parameters):
         :param objective: Objective function to be used.
         :param parameters: List of :class:`~symfit.core.argument.Parameter` instances
         """
-        self.objective = objective
-        self.params = parameters
+        self._parameters = parameters
```
tBuLi: why the underscore here? For fixed_params I would agree, but I don't see why for parameters itself.

tests/test_general.py
```diff
@@ -844,6 +844,24 @@ def test_error_analytical(self):
...
+        a, b, c, d = parameters('a, b, c, d')
+        x, y = variables('x, y')
+        
+        c.value = 4.0
+        c.fixed = True
+        
+        model_dict = {y: a * exp(-(x - b)**2 / (2 * c**2)) + d}
+        fit = Fit(model_dict, x=xdata, y=ydata)
+        fit_result = fit.execute()
+        self.assertEqual(4.0, fit_result.params['c'])
```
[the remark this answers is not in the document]
pckroon: I'm not sure they should be the same, to be honest. I would argue that setting a variable as fixed it's just not allowed to change during the fitting (i.e. treat it as a constraint). This means that the models where there's a parameter less is fundamentally different.

Furthermore, I did not touch anything around covariance matrices, or in fit.py, so even if we leave that for now it doesn't change existing behaviour.

symfit/core/minimizers.py
pckroon: Meh. Could be just `self.parameters`. I considered requiring it down the line an implementation detail at the time.

symfit/core/minimizers.py
```diff
@@ -64,8 +67,21 @@ class GradientMinimizer(BaseMinimizer):
     def __init__(self, *args, **kwargs):
         jacobian = kwargs.pop('jacobian')
         super(GradientMinimizer, self).__init__(*args, **kwargs)
-        self.jacobian = jacobian
-
+        def _jacobian(*args, **kwargs):
+            # Assume the order of what jacobian returns and parameters is the same
```
tBuLi: This assumption is indeed guaranteed, parameters are sorted by their str representation and this is used when deriving the jacobian.

symfit/core/minimizers.py
```diff
@@ -64,8 +67,21 @@ class GradientMinimizer(BaseMinimizer):
...
+            # We need to remove the values for the fixed parameters from what
+            # jacobian returns
+            fixed_vals = {p.name: p.value for p in self._fixed_params}
+            out = jacobian(*args, **kwargs, **fixed_vals)
```
tBuLi: works only in > py3

symfit/core/minimizers.py
```diff
@@ -64,8 +67,21 @@ class GradientMinimizer(BaseMinimizer):
...
+            jac = []
+            for param, val in zip(self._parameters, out):
```
tBuLi: since you forced jacobian to be a numpy array (why?) it's cleaner to use numpy masking:
```python
mask = [p.fixed for p in self.parameters]
jac = out[mask]
```

pckroon: It's only forced to np.array in the scipy minimizer wrapper. Otherwise the scipy minimizer craps out.

tests/test_general.py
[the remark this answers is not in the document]
tBuLi: Actually you are right, I was still thinking about 0.3.x, where the covariance matrix returned by the minimizer is used. Since we now redetermine it in a uniform way my critique does not apply. Nice.

symfit/core/minimizers.py
```diff
@@ -64,8 +67,22 @@ class GradientMinimizer(BaseMinimizer):
...
+        def _jacobian(*args, **kwargs):
+            # Assume the order of what jacobian returns and parameters is the same
+            # We need to remove the values for the fixed parameters from what
+            # jacobian returns
+            fixed_vals = {p.name: p.value for p in self._fixed_params}
+            kwargs.update(fixed_vals)
+            out = jacobian(*args, **kwargs)
+            jac = []
+            for param, val in zip(self.parameters, out):
+                if not param.fixed:
```
tBuLi: since you forced the jacobian to be a numpy array (why?) it's cleaner to use numpy masking:
```python
mask = [p.fixed for p in self.parameters]
jac = out[mask]
```

pckroon: Because it's not definitely a np array here. The scipy wrapper makes it's objective and jacobian return numpy arrays, to appease the minimizer gods.

PR #153, opened 2018-05-06
Fix #42 and #82.

### LINKED ISSUE #42: Docs: add glossary
I would love for the docs to have a glossary of commonly used (statistical) terms and their precise definition in symfit.

When working with a certain fitting framework commonly used in particle physics I noticed that they use the term error both when fitting and when making error bar plots. However, in one context one assumes this to mean standard deviation, in the other standard error. 

Such ambiguities should be prevented and a glossary might go a long way in preventing them.

### LINKED ISSUE #82: self.sigmas identical to self.sigma_data
These two seem to contain the same information (on `Fit` objects). To prevent future bugs from popping up, this needs to be checked more carefully and if so, only self.sigma_data should be kept.

tBuLi:
data dicts on `TakesData`-children now have Symbols as dict keys, no longer just the names.
This also fixed #82 in the process.

### DIFF
--- a/symfit/contrib/interactive_guess/tests/test_interactive_fit.py
+++ b/symfit/contrib/interactive_guess/tests/test_interactive_fit.py
@@ -58,7 +58,8 @@ class Gaussian2DInteractiveGuessTest(unittest.TestCase):
             self.guess._sliders[self.guess.model.params[0]].reset()
 
     def test_slider_callback_data(self):
-        x_points = self.guess._x_points[self.guess.model.independent_vars[0]]
+        x = self.guess.model.independent_vars[0]
+        x_points = self.guess._x_points[x]
         hi = np.max(x_points)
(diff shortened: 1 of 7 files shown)


pckroon:
Looks much better :)
Carries my seal of approval.

symfit/contrib/interactive_guess/tests/test_interactive_fit.py
```diff
@@ -77,14 +77,14 @@ def test_slider_callback_data(self):
     def test_get_data(self):
         y = self.guess.model.dependent_vars[0]
         x = self.guess.model.independent_vars[0]
-        x_points = self.guess._x_points['x']
+        x_points = self.guess._x_points[self.guess.model.independent_vars[0]]
```
tBuLi: use x from the line above

PR #157, opened 2018-05-10
Fixed #135. ODEModels can now be evaluated for any t-value.

### LINKED ISSUE #135: ODEModels at t < t_initial
Extend `ODEModel`'s so they can also be evaluated for `t < t_initial`. In principle this is easy to add; it just wasn't a priority when I first implemented them.

tBuLi:
ODEModels can now be evaluated for any t-value, integrating back in time from the initial values has now works seamlessly.

### DIFF
--- a/symfit/core/fit.py
+++ b/symfit/core/fit.py
@@ -1753,33 +1753,53 @@ class ODEModel(CallableModel):
         Dfun = lambda ys, t, *a: [[c(t, *(list(ys) + list(a))) for c in row] for row in self._njacobian]
 
         initial_dependent = [self.initial[var] for var in self.dependent_vars]
-        initial_independent = self.initial[self.independent_vars[0]] # Assuming there's only one
+        t_initial = self.initial[self.independent_vars[0]] # Assuming there's only one
 
         # Check if the time-like data includes the initial value, because integration should start there.
-        # For fitting to make sence, it should probably not be in there though. Needs mathematical backing.
         try:
             t_like[0]
         except (TypeError, IndexError): # Python scalar gives TypeError, numpy scalars IndexError
             t_like = np.array([t_like]) # Allow evaluation at one point.
 
-        if t_like[0] == initial_independent:
-            start = 0
-            warnings.warn("The initial point should probably not be included with your data points as this point will always be fitted perfectly.")
-        elif t_like[0] < initial_independent:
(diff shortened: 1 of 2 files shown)


symfit/core/fit.py
```diff
@@ -1753,33 +1749,53 @@ def eval_components(self, *args, **kwargs):
...
-            t_like = np.hstack((np.array([initial_independent]), t_like))
-            start = 1
-        ans = odeint(
+            t_bigger = np.concatenate(
+                (np.array([t_initial]), t_like[t_like > t_initial])
+            )
+            t_smaller = np.concatenate(
+                (np.array([t_initial]), np.flip(t_like[t_like < t_initial], axis=-1))
+            )
+            t_total = np.concatenate((t_like[t_like < t_initial] ,t_bigger))
```
pckroon: Why not `t_total = np.concatenate((t_smaller[:-1], t_bigger))`?
And this line should probably be dedented one tick.

symfit/core/fit.py
```diff
@@ -1753,33 +1749,53 @@ def eval_components(self, *args, **kwargs):
...
+                bound_arguments.arguments[param.name] for param in self.params),
+            Dfun=Dfun,
+            *self.lsoda_args, **self.lsoda_kwargs
+        )
+
+        ans = np.concatenate((np.flip(ans_smaller[1:], axis=0), ans_bigger))
+        if t_initial in t_like:
+            return ans.T
+        else:
+            return ans[t_total != t_initial].T
```
pckroon: This needs a comment I think

tBuLi: Good point

symfit/core/fit.py
tBuLi: changed it to `t_total = np.concatenate((t_smaller[::-1][:-1], t_bigger))`

symfit/core/fit.py
```diff
@@ -1770,7 +1771,8 @@ def eval_components(self, *args, **kwargs):
             t_smaller = np.concatenate(
                 (np.array([t_initial]), np.flip(t_like[t_like < t_initial], axis=-1))
             )
-            t_total = np.concatenate((t_like[t_like < t_initial] ,t_bigger))
+        # Properly ordered time axis containing t_initial
+        t_total = np.concatenate((t_smaller[::-1][:-1], t_bigger))
```
pckroon: Why `[::-1]` here, but `np.flip` everywhere else?

PR #158, opened 2018-05-18
Make parameters fully pickleable

pckroon:
It should be noted that the `__getstate__` from the sympy parent classes is wrong, so I made a new one. Symbol.name survives pickling via the __getnewargs__ method. We can't use that mechanism, since we need them to be passed to __init__.

### DIFF
--- a/symfit/core/argument.py
+++ b/symfit/core/argument.py
@@ -1,7 +1,13 @@
+from collections import defaultdict
 import numbers
 import warnings
 
 from sympy.core.symbol import Symbol
+
+# This can not be a class attribute, since those become read only when they're
+# __slots__. Instead, make it a global dictionary of {class: count}.
+argument_indices = defaultdict(int)
+
 
 class Argument(Symbol):
     """
@@ -20,6 +26,8 @@ class Argument(Symbol):
         print(y.name)
         >> 'y'
     """
+    __slots__ = ['_argument_index', '_argument_name']
+
     def __new__(cls, name=None, *args, **assumptions):
         assumptions['real'] = True
         # Generate a dummy name
@@ -32,13 +40,13 @@ class Argument(Symbol):
                 DeprecationWarning, stacklevel=2
             )
 
-            name = '{}_{}'.format(cls._argument_name, cls._argument_index)
+            name = '{}_{}'.format(cls._argument_name, argument_indices[cls])
             instance = super(Argument, cls).__new__(cls, name, **assumptions)
-            instance._argument_index = cls._argument_index
-            cls._argument_index += 1
-            return instance
         else:
-            return super(Argument, cls).__new__(cls, name, **assumptions)
+            instance = super(Argument, cls).__new__(cls, name, **assumptions)
+        instance._argument_index = argument_indices[cls]
+        argument_indices[cls] += 1
+        return instance
 
     def __init__(self, name=None, *args, **assumptions):
         # TODO: A more careful look at Symbol.__init__ is needed! However, it
@@ -66,9 +74,9 @@ class Parameter(Argument):
     be generated.
     """
     # Parameter index to be assigned to generated nameless parameters
-    _argument_index = 0
+    __slots__ = ['min', 'max', 'fixed', 'value']
+
     _argument_name = 'par'
-    __slots__ = ['min', 'max', 'fixed']
 
     def __new__(cls, name=None, *args, **kwargs):
         try:
@@ -106,5 +114,5 @@ class Parameter(Argument):
 class Variable(Argument):
     """ Variable type."""
     # Variable index to be assigned to generated nameless variables
-    _argument_index = 0
     _argument_name = 'var'
+    __slots__ = ()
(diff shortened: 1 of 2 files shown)


symfit/core/argument.py
```diff
@@ -58,6 +68,7 @@ class Parameter(Argument):
     # Parameter index to be assigned to generated nameless parameters
     _argument_index = 0
     _argument_name = 'par'
+    __slots__ = ['min', 'max', 'fixed']
```
tBuLi: Don't forget 'name' like symbols have too.

pckroon: Nope, don't have to. I just define the slots that *my* class needs, and I don't have to care about my parents. 
See https://stackoverflow.com/questions/1816483/python-how-does-inheritance-of-slots-in-subclasses-actually-work

tBuLi: Nice, that's good to know.

tBuLi:
I'm a bit surprised about the sympy getstate not being sufficient, because I think they are pickelable right? Do you have an explanation for this?

symfit/core/argument.py
```diff
@@ -46,6 +47,15 @@ def __init__(self, name=None, *args, **assumptions):
             self.name = name
         super(Argument, self).__init__()
 
+    def __getstate__(self):
+        state = super(Argument, self).__getstate__()
+        state.update(dict(
+            (slot, getattr(self, slot))
```
tBuLi: Instead of this dict I would prefer to use the dict comprehension syntax, as it is more consistent with the rest of the project.

pckroon:
If you have `__slots__` you must also have a `__gestate__` (and a `__setstate__`, bu there the Sympy's works: it just updates all the attributes after creation). `Symbol.__getstate__` looks like this:
```python
    def __getstate__(self):
        return {'_assumptions': self._assumptions}
```
Which of course is fine, if and only if you only have `_assumptions`. The rest of it's magice (`name`) has to be provided on object creation (not `__init__`, but `__new__`) which means this mechanism won't work. Instead, they're dealt with by `__getnewargs__`.

tests/test_argument.py
```diff
@@ -75,6 +76,36 @@ def test_symbol_add(self):
...
+        self.assertEqual((A.min, A.value, A.max, A.fixed, A.name),
+                         (new_A.min, new_A.value, new_A.max, new_A.fixed, new_A.name))
+
+    def test_slots(self):
+        """
+        Make sure Parameters and Variables don't have a __dict__
+        """
+        P = Parameter('P')
+        
+        print([x.__slots__ for x in Parameter.__mro__ if hasattr(x, '__slots__')])
```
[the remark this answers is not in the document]
pckroon: Mwah. Pytest captures and trashes output anyway, unless the test fails.

tBuLi:
Thanks for the explanation. 

And are you sure that the class attributes are basically broken by __slots__, and this global dict is the best workaround? If so, I will merge this.

tBuLi:
whoops wrong button, premature close

pckroon:
As it was it breaks on me saying that attribute _argument_index is read-only. This is I think the only workaround I saw. Maybe an alternative would be something like making a meta-class, or a singleton object that keeps track of this dict. Either way, overly complicated.

tBuLi:
That is indeed a bit troubling, but this is a deprecated feature anyway so I'm not sure how much time should be spend on it. 

So I don't think pickling issues with automatically generated names are a big problem, since they are now against best practices anyway, so it's not unreasonable to demand people to be follow best practices when using multiprocessing for example. Would it be possible to modify `__setstate__` to set `argument_indices[cls] = max(argument_indices[cls], current_index)`?

My biggest problem is that I want to avoid globals if in any way possible, so I would prefer any other solution over this but if there is no other way then we will do this since this feature is deprecated anyway.

tBuLi:
I agree with that.

But I just looked at the source of sympy.Dummy objects and they don't seem to have a problem with the class attribute, but it has to be a different one. See http://docs.sympy.org/1.0/modules/core.html#sympy.core.symbol.Dummy.

Mimicking this is definitely the preferred solution.

pckroon:
But that does mean your object will have a `__dict__`, negating some (all?) of the benefits involving `__slots__`

tBuLi:
I finally had some time to check but `Dummy`'s don't have `__dict__`'s, only `__slots__`. So I would prefer mimicking their solution to using a global variable.

tests/test_argument.py
```diff
@@ -75,6 +76,36 @@ def test_symbol_add(self):
         new = x + y
         self.assertIsInstance(new, sympy.Add)
 
+    def test_pickle(self):
+        """
+        Make sure attributes are preserved when pickling
+        """
+        A = Parameter('A', min=0., max=1e3, fixed=True)
+        new_A = pickle.loads(pickle.dumps(A))
```
tBuLi: Perhaps you can add the same test for automatically generated names for as much as possible?

pckroon: Do we want to test against deprecated behaviour?

tBuLi: It's depricated, but still one of the futures now right? So it should be pickalable, no?

pckroon: Whatever floats your boat ;)

PR #163, opened 2018-07-04
Fixed #161.

### LINKED ISSUE #161: shape mismatch for sum
I try to run the following snippet but get an exception:

```python
import symfit

temperature = symfit.Variable('temperature')
length = symfit.Variable('length')
shift = symfit.Parameter('s', value=300)
model = {length: shift + 1 / (1 + symfit.exp(- temperature))}

temp_data = [270, 280, 285, 290, 295, 300, 310, 320]
length_data = [8.33, 8.41, 8.45, 8.5, 8.54, 9.13, 9.27, 9.4]
fit = symfit.Fit(model, length=length_data, temperature=temp_data)
fir_result = fit.execute()
```

The exception:

```
ValueError                                Traceback (most recent call last)
<ipython-input-4-47e2b6349ab2> in <module>()
      9 length_data = [8.33, 8.41, 8.45, 8.5, 8.54, 9.13, 9.27, 9.4]
     10 fit = symfit.Fit(model, length=length_data, temperature=temp_data)
---> 11 fir_result = fit.execute()

~/dev/symfit/symfit/core/fit.py in execute(self, **minimize_options)
   1389         else:
   1390             if cov_matrix is None:
-> 1391                 cov_matrix = self.covariance_matrix(dict(zip(self.model.params, minimizer_ans._popt)))
   1392         finally:
   1393             minimizer_ans.covariance_matrix = cov_matrix

~/dev/symfit/symfit/core/fit.py in covariance_matrix(self, best_fit_params)
    873             if len(set(arr.shape for arr in self.sigma_data.values())) == 1:
    874                 # Shapes of all sigma data identical
--> 875                 return self._cov_mat_equal_lenghts(best_fit_params=best_fit_params)
    876             else:
    877                 return self._cov_mat_unequal_lenghts(best_fit_params=best_fit_params)

~/dev/symfit/symfit/core/fit.py in _cov_mat_equal_lenghts(self, best_fit_params)
    951             jac = jac * np.ones_like(W)
    952         # Dot away all but the parameter dimension!
--> 953         cov_matrix_inv = np.tensordot(W*jac, jac, (range(1, jac.ndim), range(1, jac.ndim)))
    954         cov_matrix = np.linalg.inv(cov_matrix_inv)
    955         return cov_matrix

/coarse/jon/Envs/apricot/lib/python3.6/site-packages/numpy/core/numeric.py in tensordot(a, b, axes)
   1281                 axes_b[k] += ndb
   1282     if not equal:
-> 1283         raise ValueError("shape-mismatch for sum")
   1284 
   1285     # Move the axes to sum over to the end of "a"

ValueError: shape-mismatch for sum
```

According to `pip show` I am using version  0.3.3.dev487 of symfit, version 1.1.1 of sympy, and 1.14.3 of numpy.

I try to simplify the model to `{length: shift - temperature}` and I get the same error.

tBuLi:
In case of failure to estimate the shape of the jacobian, we check if this is becuase the shape of the data-dimension is not properly estimated and adjust if necessary.

### DIFF
--- a/symfit/core/fit.py
+++ b/symfit/core/fit.py
@@ -842,7 +842,7 @@ class BaseFit(TakesData):
         raise NotImplementedError('Every subclass of BaseFit must have an eval_jacobian method.')
 
 
-class HasCovarianceMatrix(object):
+class HasCovarianceMatrix(TakesData):
     """
     Mixin class for calculating the covariance matrix for any model that has a
     well-defined Jacobian :math:`J`. The covariance is then approximated as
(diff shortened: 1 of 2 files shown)


tBuLi:
Yes, they failed without the fix. I agree that these tests are not great, but I struggle to come up with a better way to test for this.

I thought about that, but since there is actually a lot of `self.dependent_data`-like statements in `HasCovarianceMatrix`, it already assumed that it has data. So I think this way also takes care of the implicit dependency there.

Taking the shape from W itself is not a bad idea, but I went this way because of the argument above. Do you agree with the rationale?

pckroon:
Yes, makes sense :) Nice work

PR #167, opened 2018-07-09
Add __version__

tBuLi:
Extracts the version number using `pkg_resources` and assigns it to `symfit.__version__`.

This is then used to automatically give the correct version number to the docs.


pckroon:
Looks fine from here. Does it actually work and give a sane version?

tBuLi:
Based on [this](https://docs.readthedocs.io/en/latest/versions.html), I made sure that readthedocs builds from the latest stable version. This means it takes the most recent tag, e.g. 0.4.4, pip installs that, and then proceeds to build the docs. 

I have additionally checked that the docs for this PR-branch can be build without error, though I cannot see what version number it decides on. 

But based on those two things, I think we can take a leap of faith and assume that it is able to extract a version from pip, and probably the correct one.

Issue #165, opened 2018-07-07
Constrained Example Error


bhclowers:
As of version 0.4.2 the example provided on the front of the github page for constraints appears to throw and error (at least on my system Python 3.5.2).  Any ideas would be appreciated. 

```
from symfit import parameters, Fit, Equality, GreaterThan
x, y = parameters('x, y')
model = 2 * x * y + 2 * x - x**2 - 2 * y**2
constraints = [
    Equality(x**3, y),
    GreaterThan(y, 1),
]

fit = Fit(model, constraints=constraints)
fit_result = fit.execute()
```
Error output:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-5bb7e724f8a2> in <module>()
      9 
     10 fit = Fit(model, constraints=constraints)
---> 11 fit_result = fit.execute()

~/.local/lib/python3.5/site-packages/symfit/core/fit.py in execute(self, **minimize_options)
   1401         :return: FitResults instance
   1402         """
-> 1403         minimizer_ans = self.minimizer.execute(**minimize_options)
   1404         try: # to build covariance matrix
   1405             cov_matrix = minimizer_ans.covariance_matrix
~/.local/lib/python3.5/site-packages/symfit/core/minimizers.py in execute(self, **minimize_options)
    207             method='SLSQP',
    208             bounds=self.bounds,
--> 209             **minimize_options
    210         )
    211 
~/.local/lib/python3.5/site-packages/symfit/core/minimizers.py in execute(self, **minimize_options)
    194 
    195     def execute(self, **minimize_options):
--> 196         return super(ScipyGradientMinimize, self).execute(jacobian=self.wrapped_jacobian, **minimize_options)
    197 
    198 
~/.local/lib/python3.5/site-packages/symfit/core/support.py in wrapped_func(*args, **kwargs)
    280                     else:
    281                         bound_args.arguments[param.name] = param.default
--> 282             return func(*bound_args.args, **bound_args.kwargs)
    283         return wrapped_func
    284 
~/.local/lib/python3.5/site-packages/symfit/core/minimizers.py in execute(self, bounds, jacobian, **minimize_options)
    129             constraints=self.constraints,
    130             jac=jacobian,
--> 131             **minimize_options
    132         )
    133         # Build infodic
[...]

PR #168, opened 2018-07-12
Fixed params and constraints (#165)

tBuLi:
This is a first attempt at fixing #165.

To do:

- [x] Fixing the problem.
- [x] Add tests
- [x] Add better docstring to all `ScipyMinimize` base classes.
- [x] Make `ScipyConstrainedMinimize` not inheret from `ScipyGradientMinimize` because this only applies to SLSQP, not to COBYLA 
- [x] Use the @wraps decorator from functools on all the wrapping functions we use in `minimizers.py`. This makes tracebacks and inspect a lot clearer.
- [x] Rename `ScipyMinimize.wrap_func` to something more descriptive like ~~`positional2keyword`~~ `list2keyword`
- [x] update `resize_jac` to use a mask and numpy indexing instead of a list such that it still returns an array instead of a list.


symfit/core/minimizers.py
```diff
@@ -172,43 +193,70 @@ def scipy_constraints(constraints, data):
...
 
 
-class SLSQP(ScipyGradientMinimize, ConstrainedMinimizer, BoundedMinimizer):
+class SLSQP(ScipyConstrainedMinimize, GradientMinimizer, BoundedMinimizer):
+    def __init__(self, *args, **kwargs):
+        super(SLSQP, self).__init__(*args, **kwargs)
+        # We have to break DRY because you cannot inherit from both
+        # ScipyConstrainedMinimize and ScipyGradientMinimize. So SLQSP is a
+        # special case. This is the same code as in ScipyGradientMinimize.
+        self.wrapped_jacobian = self.list2kwargs(self.wrapped_jacobian)
```
pckroon: That mean we did *something* wrong in class design. I don't really see where though.
Probably `ScipyMinimize` has to actually do something, and `ScipyGradient` and `ScipyConstraint` need to be purely mix-ins that can't exist on themselves. I suggest we leave it for now.

pckroon:
If I were you I'd `git rebase -i origin/master` to resolve the merge conflict, and then `git push -f` to this branch again. But it's up to you ;)

symfit/core/minimizers.py
tBuLi: Yeah, that was also my thinking. They should probably be made into pure mixin classes, but since SLSQP is the only one breaking the pattern we can treat it like a special snowflake for now ;).

pckroon: We might want to expand the comment with this. Just so we don't have to come up with this again

symfit/core/fit.py
```diff
@@ -12,7 +11,8 @@
 from scipy.integrate import odeint
 
 from symfit.core.argument import Parameter, Variable
-from .support import seperate_symbols, keywordonly, sympy_to_py, cache, key2str
+from .support import \
+    seperate_symbols, keywordonly, sympy_to_py, cache, key2str, partial
```
pckroon: Why not import repeatable_partial (as partial)? This way we still end up with the partial from functools I think

symfit/core/minimizers.py
```diff
@@ -1,20 +1,21 @@
...
 import sys
 from collections import namedtuple, Counter
-from functools import partial
 
 from scipy.optimize import minimize, differential_evolution
 import sympy
 import numpy as np
 
-from .support import key2str, keywordonly
+from .support import key2str, keywordonly, partial
```
pckroon: Same as previous comment

symfit/core/support.py
```diff
@@ -21,8 +20,15 @@
...
     import inspect as inspect_sig
+    from functools import wraps
 else:
     import funcsigs as inspect_sig
+    from functools32 import wraps
+
+if sys.version_info >= (3, 5):
+    from functools import partial
+else:
+    from ._support import repeatable_partial as partial
```
pckroon: The file _support needs to be renamed if it's just this partial hack. Currently it's hard to track what's coming from where.

symfit/core/fit.py
tBuLi: In `support.py`, I check for the python version. If this is py3.5 or later, `partial` is defined as `functools.partial`. Otherwise, this gives you `repeatable_partial` as `partial`.

symfit/core/support.py
tBuLi: Perhaps it could be named `_repeatable_partial.py` instead? The reason for having it separate is such that `repeatable_partial` is only ever defined in relevant python versions.

pckroon: That works for me.

pckroon:
Looks good to me. Travis had a hickup on the `pr` build and works when restarted.

Merge at will!

Issue #171, opened 2018-07-13
sympy 1.2 compatibility issue


tBuLi:
Related to #169, `sympy == 1.2` seems to introduce a big problem for us because in `printer.py` `sympy` takes `printmethod = '_print_' + cls.__name__` in order to retrieve the proper way to print the different types of object they have.

The problem is that there now seems to be a function called `_print_Variable`, whereas `symfit.Variable` objects inherit from `sympy.Symbol` and should therefore be printed using `_print_Symbol`.

We should therefore take the following steps:
- [x] Investigate if indeed this function `_print_Variable` did not exist before 1.2
- [x] Raise the issue in the sympy issue tracker

tBuLi:
Indeed, `_print_Variable` does not yet exist in `sympy 1.1.1`, as do some other functions in `codeprinter.py`.

It would seem that objects called `Variable` exist in certain submodules of sympy, though they have existed for longer already. https://github.com/sympy/sympy/search?q=%22class+Variable%22&unscoped_q=%22class+Variable%22

Looking at the line in `_print_Variable`, it seems to relate to this object: https://github.com/sympy/sympy/blob/03268800330a27da9e43ebba9eb58c0b00683261/sympy/codegen/ast.py#L1314

tBuLi:
The discussion can be followed at https://github.com/sympy/sympy/issues/14911

PR #172, opened 2018-07-15
Fix 169/171

tBuLi:
Solves the issue by instructing `sympy`'s printer on how this object should be printed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 numpy >= 1.12
 scipy >= 1.0
-sympy <= 1.1.1
+sympy
 funcsigs; python_version < '3.0'
 functools32; python_version < '3.0'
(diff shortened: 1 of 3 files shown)


tBuLi:
Just making objects tell sympy how they would like to be printed does not seem to be enough yet with sympy 1.2, follow the discussion on sympy's issue tracker linked in #171 for more details.

pckroon:
This PR also makes Travis install sympy 1.1, making it misleading. Also change the requirements in this PR.

tBuLi:
Cleaned up the commit history somewhat.

tBuLi:
Actually I just found out that there is also still the same problem in sympy, which I had fixed but forgotten about :P. So it turns out that this change is indeed needed on our end, but I also have to make a PR for sympy.

tBuLi:
Solved with #199

PR #177, opened 2018-09-04
BasinHopping -> wraps scipy.optimize.basinhopping

tBuLi:
This PR wraps `scipy.optimize.basinhopping` into a friendly `symfit` minimizer. When used through `Fit` (recommended) the full feature set of `symfit` is easily available: `symfit` provides the jacobian automatically,and we have automatic support of bounds and constraints through the standard API.

To do:

- [x] check the performance. I noticed that the direct call to `scipy.optimize.basinhopping` and `BasinHopping` are fast, but when done through `Fit` the performance seems significantly worse. (From answering directly to having to wait a couple of seconds) It has to be investigated weather this is due to overhead, or that every evaluation is more expensive because of all the wrapping and partialing functions `symfit` does this way.
- [x] Write docs.
- [x] Use `cls.minimize_method` by default to determine the method of `ScipyMinimize`'s children, to reduce boiler plate.

If this is suitable to large numbers of parameters then I intend to use this as the default solver for future `IndexedParameter`'s if no quick-fire solution is known. If not, we might need to wrap a true monte-carlo package to do that.

### DIFF
--- a/docs/fitting_types.rst
+++ b/docs/fitting_types.rst
@@ -143,6 +143,7 @@ Example::
 ``fit_result`` is a normal :class:`~symfit.core.fit_results.FitResults` object.
 As always, bounds on parameters and even constraints are supported.
 
+.. _minimize_maximize:
 Minimize/Maximize
 -----------------
 Minimize or Maximize a model subject to bounds and/or constraints. As an example
@@ -459,6 +460,58 @@ to your answer, and then polish it off using a local minimizer::
   There is no way to guarantee that the minimum found is actually the global
   minimum. Unfortunately there is no way around this. Therefore, you should
   always critically inspect the results.
+
+Constrained Basin-Hopping
+-------------------------
+Worthy of special mention is the ease with which constraints or bounds can be
+added to :class:`symfit.core.minimizers.BasinHopping` when used through the
+:class:`symfit.core.fit.Fit` interface. As a very simple example, we shall
+compare to an example from the :mod:`scipy` docs::
+
+    import numpy as np
+    from scipy.optimize import basinhopping
+
+    def func2d(x):
+        f = np.cos(14.5 * x[0] - 0.3) + (x[1] + 0.2) * x[1] + (x[0] + 0.2) * x[0]
+        df = np.zeros(2)
+        df[0] = -14.5 * np.sin(14.5 * x[0] - 0.3) + 2. * x[0] + 0.2
+        df[1] = 2. * x[1] + 0.2
+        return f, df
+
+    minimizer_kwargs = {"method":"L-BFGS-B", "jac":True}
+    x0 = [1.0, 1.0]
+    ret = basinhopping(func2d, x0, minimizer_kwargs=minimizer_kwargs, niter=200)
+
+Let's compare to the same functionality in :mod:`symfit`
+
+    import numpy as np
+    from symfit.core.minimizers import BasinHopping
+    from symfit import parameters, Fit, cos
+
+    x0 = [1.0, 1.0]
+    x1, x2 = parameters('x1, x2', value=x0)
+
+    model = cos(14.5 * x1 - 0.3) + (x2 + 0.2) * x2 + (x1 + 0.2) * x1
+
+    fit = Fit(model, minimizer=BasinHopping)
+    fit_result = fit.execute(niter=200)
+
+No `minimizer_kwargs` have to be provided, as :mod:`symfit` will automatically
+compute and provide the jacobian and select a minimizer. In this case, `symfit`
+will choose `BFGS`. When bounds are provided, `symfit` will switch to
+using `L-BFGS-B` instead. Setting bounds is as simple as::
+
+    x1.min = 0.0
+    x1.max = 100.0
+
+However, the real strength the `symfit` syntax lies in providing constraints::
+
+    constraints = [Eq(x1, x2)]
+    fit = Fit(model, minimizer=BasinHopping, constraints=constraints)
+
+This artificial example will make sure `x1 == x2` after fitting. If you have
+read the :ref:`minimize_maximize` section, you will know how much work this
+would be in pure :mod:`scipy`.
 
 Advanced usage
 --------------
--- a/docs/fitting_types.rst
+++ b/docs/fitting_types.rst
@@ -144,6 +144,7 @@ Example::
 As always, bounds on parameters and even constraints are supported.
 
 .. _minimize_maximize:
+
 Minimize/Maximize
 -----------------
 Minimize or Maximize a model subject to bounds and/or constraints. As an example
--- a/docs/fitting_types.rst
+++ b/docs/fitting_types.rst
@@ -363,11 +363,12 @@ More common examples, such as dampened harmonic oscillators also work as expecte
 
 .. _global-fitting:
 
-Global Fitting
---------------
-In a global fitting problem, we fit to multiple datasets where one or more
-parameters might be shared. The same syntax used for ODE fitting makes this
-problem very easy to solve in :mod:`symfit`.
+Fitting multiple datasets
+-------------------------
+A common fitting problem, is to fit to multiple datasets. This is sometimes
+referred to as global fitting. In such fits parameters might be shared
+between the fits to the different datasets. The same syntax used for ODE
+fitting makes this problem very easy to solve in :mod:`symfit`.
 
 As a simple example, suppose we have two datasets measuring exponential decay,
 with the same background, but different amplitude and decay rate.
@@ -397,6 +398,10 @@ normal way::
    :width: 500px
    :alt: ODE integration
 
+Any ``Model`` that comes to mind is fair game, behind the scenes :mod:`symfit`
+will build a least squares function where the residues of all the components
+are added squared, ready to be minimized. Unlike in the above example, the
+`x`-axis does not even have to be shared between the components.
 
 .. warning::
   The regression coefficient is not properly defined for vector-valued models,
@@ -406,8 +411,8 @@ normal way::
 
   Do not cite the overall :math:`R^2` given by :mod:`symfit`.
 
-Finding the optimal solution
-----------------------------
+Global Minimization
+-------------------
 Very often, there are multiple solutions to a fitting (or minimisation)
 problem. These are local minima of the objective function. The best solution of
 course is the global minimum, but most minimization algorithms will only find a
@@ -416,10 +421,16 @@ your parameters. This can be incredibly annoying if you have no further
 knowledge about your system.
 
 Luckily, global minimizers exist which are not influenced by the initial
-guesses for your parameters. In symfit the
+guesses for your parameters. In symfit, two such algorithms from :mod:`scipy`
+have been wrapped for this pourpose. Firstly, the
 :func:`~scipy.optimize.differential_evolution` algorithm from :mod:`scipy` is
-wrapped as :class:`~symfit.core.minimizers.DifferentialEvolution`. To use it,
+wrapped as :class:`~symfit.core.minimizers.DifferentialEvolution`. Secondly,
+the :func:`~scipy.optimize.basinhopping` algorithm is available as
+:class:`~symfit.core.minimizers.BasinHopping`. To use these minimizers,
 just tell :class:`~symfit.core.fit.Fit`::
+
+    from symfit import Parameter, Variable, Model, Fit
+    from symfit.core.minimizers import DifferentialEvolution
 
     x = Parameter('x')
     x.min, x.max = -100, 100
@@ -428,8 +439,9 @@ just tell :class:`~symfit.core.fit.Fit`::
 
     model = Model({y: x**4 - 10 * x**2 - x})  # Skewed Mexican hat
     fit = Fit(model, minimizer=DifferentialEvolution)
+    fit_result = fit.execute()
 
-However, due to how the algorithm works, it's not great at finding the exact
+However, due to how this algorithm works, it's not great at finding the exact
 minimum (but it will find it if given enough time). You can work around this by
 "chaining" minimizers: first run a global minimization to (hopefully) get close
 to your answer, and then polish it off using a local minimizer::
@@ -437,14 +449,16 @@ to your answer, and then polish it off using a local minimizer::
     fit = Fit(model, minimizer=[DifferentialEvolution, BFGS])
 
 .. note::
-  Differential evolution is rather sensitive to it's hyperparameters. You might
-  need to play with those to get appropriate results::
+  Global minimizers such as differential evolution and basin-hopping are
+  rather sensitive to their hyperparameters. You might
+  need to play with those to get appropriate results, e.g.::
   
     fit.execute(DifferentialEvolution={'popsize': 20, 'recombination': 0.9})
 
 .. note::
-  There is no way to garuantee that the minimum found is actually the global
-  minimum. Unfortunately there is no way around this.
+  There is no way to guarantee that the minimum found is actually the global
+  minimum. Unfortunately there is no way around this. Therefore, you should
+  always critically inspect the results.
 
 Advanced usage
 --------------
--- a/docs/fitting_types.rst
+++ b/docs/fitting_types.rst
@@ -367,7 +367,7 @@ More common examples, such as dampened harmonic oscillators also work as expecte
 
 Fitting multiple datasets
 -------------------------
-A common fitting problem, is to fit to multiple datasets. This is sometimes
+A common fitting problem is to fit to multiple datasets. This is sometimes
 referred to as global fitting. In such fits parameters might be shared
 between the fits to the different datasets. The same syntax used for ODE
 fitting makes this problem very easy to solve in :mod:`symfit`.
@@ -400,7 +400,7 @@ normal way::
    :width: 500px
    :alt: ODE integration
 
-Any ``Model`` that comes to mind is fair game, behind the scenes :mod:`symfit`
+Any ``Model`` that comes to mind is fair game. Behind the scenes :mod:`symfit`
 will build a least squares function where the residues of all the components
 are added squared, ready to be minimized. Unlike in the above example, the
 `x`-axis does not even have to be shared between the components.
@@ -483,7 +483,7 @@ compare to an example from the :mod:`scipy` docs::
     x0 = [1.0, 1.0]
     ret = basinhopping(func2d, x0, minimizer_kwargs=minimizer_kwargs, niter=200)
 
-Let's compare to the same functionality in :mod:`symfit`
+Let's compare to the same functionality in :mod:`symfit`::
 
     import numpy as np
     from symfit.core.minimizers import BasinHopping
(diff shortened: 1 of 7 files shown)


symfit/core/minimizers.py
```diff
@@ -446,15 +456,87 @@ def execute(self, **minimize_options):
...
         return super(NelderMead, self).execute(method='Nelder-Mead', **minimize_options)
 
 
+class BasinHopping(ScipyMinimize, BaseMinimizer):
+    """
+    Wrapper around :func:`scipy.optimize.basinhopping`'s basin-hopping algorithm.
+
+    As always, the best way to use this algorithm is through
+    :class:`~symfit.core.fit.Fit`, as this will automatically select a local
+    minimizer for you depending on wether you provvided ounds, constraints, etc.
```
pckroon: Needless bit of docstring

tBuLi:
I have now cProfiled the following equivalent scripts:

Pure scipy, 0.644 seconds:
```python
import numpy as np
from scipy.optimize import basinhopping

def func2d(x):
    f = np.cos(14.5 * x[0] - 0.3) + (x[1] + 0.2) * x[1] + (x[0] + 0.2) * x[0]
    df = np.zeros(2)
    df[0] = -14.5 * np.sin(14.5 * x[0] - 0.3) + 2. * x[0] + 0.2
    df[1] = 2. * x[1] + 0.2
    return f, df

np.random.seed(555)
minimizer_kwargs = {'method': 'BFGS', 'jac': True}
x0 = [1.0, 1.0]
res = basinhopping(func2d, x0, minimizer_kwargs=minimizer_kwargs, niter=200)
```
Using the wrapped minimizer, 1.381 seconds:
```python
import numpy as np
from symfit.core.minimizers import BFGS, BasinHopping
from symfit import parameters

def func2d_symfit(x1, x2):
    f = np.cos(14.5 * x1 - 0.3) + (x2 + 0.2) * x2 + (x1 + 0.2) * x1
    return f

def jac2d_symfit(x1, x2):
    df = np.zeros(2)
    df[0] = -14.5 * np.sin(14.5 * x1 - 0.3) + 2. * x1 + 0.2
    df[1] = 2. * x2 + 0.2
    return df

x0 = [1.0, 1.0]
np.random.seed(555)
x1, x2 = parameters('x1, x2', value=x0)
fit = BasinHopping(func2d_symfit, [x1, x2], local_minimizer=BFGS)
minimizer_kwargs = {'jac': fit.list2kwargs(jac2d_symfit)}
fit_result = fit.execute(niter=200, minimizer_kwargs=minimizer_kwargs)
```
And pure symfit, 26.648 seconds:
```python
import numpy as np
from symfit.core.minimizers import BasinHopping
from symfit import parameters, Fit, cos

x0 = [1.0, 1.0]
np.random.seed(555)
x1, x2 = parameters('x1, x2', value=x0)
model = cos(14.5 * x1 - 0.3) + (x2 + 0.2) * x2 + (x1 + 0.2) * x1
fit = Fit(model, minimizer=BasinHopping)
fit_result = fit.execute(niter=200)
```
API-wise, I know which one I prefer, but the difference in performance is staggering.

Here are the 10 most expensive function calls per script:
```
Wed Sep  5 09:54:27 2018    scipy_basinhopping

         162532 function calls (159076 primitive calls) in 0.644 seconds

   Ordered by: internal time
   List reduced from 1387 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    51/40    0.089    0.002    0.110    0.003 {built-in method _imp.create_dynamic}
     1171    0.083    0.000    0.083    0.000 {built-in method nt.stat}
      201    0.050    0.000    0.190    0.001 optimize.py:904(_minimize_bfgs)
       13    0.045    0.003    0.045    0.003 {built-in method builtins.compile}
     2125    0.025    0.000    0.028    0.000 profile_scipy_basinhopping.py:4(func2d)
      254    0.024    0.000    0.024    0.000 {built-in method marshal.loads}
      266    0.019    0.000    0.031    0.000 <frozen importlib._bootstrap_external>:830(get_data)
     3848    0.017    0.000    0.072    0.000 optimize.py:290(function_wrapper)
     1151    0.016    0.000    0.100    0.000 linesearch.py:106(scalar_search_wolfe1)
     3276    0.013    0.000    0.013    0.000 {method 'reduce' of 'numpy.ufunc' objects}

```

```
Wed Sep  5 09:54:41 2018    scipy_basinhopping_wrapper

         619982 function calls (610008 primitive calls) in 1.381 seconds

   Ordered by: internal time
   List reduced from 3066 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     2607    0.187    0.000    0.187    0.000 {built-in method nt.stat}
    52/41    0.091    0.002    0.112    0.003 {built-in method _imp.create_dynamic}
      639    0.082    0.000    0.082    0.000 {built-in method marshal.loads}
       16    0.064    0.004    0.064    0.004 {built-in method builtins.compile}
    68862    0.050    0.000    0.050    0.000 {built-in method builtins.hasattr}
      201    0.049    0.000    0.320    0.002 optimize.py:904(_minimize_bfgs)
      657    0.048    0.000    0.077    0.000 <frozen importlib._bootstrap_external>:830(get_data)
1363/1360    0.036    0.000    0.179    0.000 {built-in method builtins.__build_class__}
      390    0.032    0.000    0.081    0.000 assumptions.py:315(__init__)
      657    0.029    0.000    0.029    0.000 {method 'read' of '_io.FileIO' objects}
```

```
Wed Sep  5 09:55:27 2018    symfit_basinhopping

         21846460 function calls (19870896 primitive calls) in 26.648 seconds

   Ordered by: internal time
   List reduced from 3477 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
3138560/2083022    1.826    0.000    9.624    0.000 {built-in method builtins.hasattr}
242514/57745    0.925    0.000   17.783    0.000 printer.py:235(_print)
424204/416475    0.891    0.000    1.409    0.000 sympify.py:54(sympify)
833503/830562    0.865    0.000    2.445    0.000 <frozen importlib._bootstrap>:997(_handle_fromlist)
   2644/1    0.824    0.000   26.811   26.811 {built-in method builtins.exec}
    19259    0.771    0.000    0.878    0.000 {built-in method builtins.eval}
   136668    0.738    0.000    3.105    0.000 evalf.py:1338(evalf)
   354136    0.680    0.000    4.263    0.000 precedence.py:113(precedence)
19252/19249    0.666    0.000    6.068    0.000 expr.py:965(as_terms)
  2240804    0.655    0.000    0.657    0.000 {built-in method builtins.isinstance}
```

First of all, it really does seem like asking for forgiveness is better than asking for permission, those hasattr calls have to go. Second, a lot of time is spend in the `sympy` printer. This has never been this noticeable before. Now, my suspicion here is that the conversion to scipy friendly functions is performed multiple times here.

More research is needed when I have time, but my first instinct would be to re-enable the `@cache` decorator on `numerical_components` on `Model`'s, such that the conversion is done only once per model. However, this does mean that if someone updates the `model_dict` after the first call to `numerical_components`, those changes will not be reflected. A solution could be to turn `model_dict` into a property so this updating could be performed.

symfit/core/minimizers.py
```diff
@@ -316,6 +316,16 @@ def _pack_output(self, ans):
                 fit_results['hessian_inv'] = ans.hess_inv
         return FitResults(**fit_results)
 
+    @classmethod
+    def minimize_method(cls):
```
pckroon: `minimizer_name`? `name`? `algorithm_name`?

symfit/core/minimizers.py
```diff
@@ -446,15 +456,87 @@ def execute(self, **minimize_options):
...
 class NelderMead(ScipyMinimize, BaseMinimizer):
     """
     Wrapper around :func:`scipy.optimize.minimize`'s NelderMead algorithm.
     """
+    @classmethod
+    def minimize_method(cls):
+        return 'Nelder-Mead'
+
     def execute(self, **minimize_options):
         return super(NelderMead, self).execute(method='Nelder-Mead', **minimize_options)
```
pckroon: Use the shiny new `minimize_method` ;)

symfit/core/minimizers.py
```diff
@@ -446,15 +456,87 @@ def execute(self, **minimize_options):
...
+    """
+    @keywordonly(local_minimizer=BFGS)
+    def __init__(self, *args, **kwargs):
+        self.local_minimizer = kwargs.pop('local_minimizer')
+        super(BasinHopping, self).__init__(*args, **kwargs)
+
+        type_error = TypeError(
+            'Currently only subclasses of ScipyMinimize are supported, since'
+            ' `scipy.optimize.basinhopping` uses `scipy.optimize.minimize`.'
+        )
```
pckroon: I wouldn't do this. Make a `message` variable, and reuse that.

symfit/core/minimizers.py
```diff
@@ -446,15 +456,87 @@ def execute(self, **minimize_options):
...
+        try:
+            issubclass(self.local_minimizer, ScipyMinimize)
+        except TypeError:
+            # It is not a class
+            if not isinstance(self.local_minimizer, ScipyMinimize):
+                raise type_error
```
pckroon: else? This decision tree needs some more work/attention and comments

symfit/core/support.py
```diff
@@ -135,12 +137,49 @@ def parameters(names, **kwargs):
...
+        `value`, `min` and `max` will be handled separately if they are sequences.
     :return: iterable of :class:`symfit.core.argument.Parameter` objects
     """
-    return symbols(names, cls=Parameter, seq=True, **kwargs)
+    sequence_fields = ['value', 'min', 'max', 'fixed']
+    sequences = {}
+    for attr in sequence_fields:
+        # if attr in kwargs:
+        try:
+            len(kwargs[attr])
```
pckroon: Should be: `iter(kwargs[attr])`

symfit/core/support.py
```diff
@@ -135,12 +137,49 @@ def parameters(names, **kwargs):
...
+        except (TypeError, KeyError):
+            # Not iterable
+            pass
+        else:
+            sequences[attr] = kwargs.pop(attr)
+
+    if 'min' in sequences and 'max' in sequences:
+        for min, max in zip(sequences['min'], sequences['max']):
+            if min > max:
```
pckroon: min and max are bad variable names.

symfit/core/support.py
```diff
@@ -135,12 +137,49 @@ def parameters(names, **kwargs):
...
+                raise ValueError('The value of `min` should be less than or'
+                                 ' equal to the value of `max`.')
+
+    ans = symbols(names, cls=Parameter, seq=True, **kwargs)
+    for key, values in sequences.items():
+        if not isinstance(values, repeat) and len(values) != len(ans):
+            raise ValueError(
+                '`len` of keyword-argument `{}` does not match the number of '
+                '`Parameter`s created.'.format(attr)
+            )
```
pckroon: I wouldn't test for this case I think. `values` might be iterators/generators/whatevers that don't have a `len`.
Instead, consider the following alternative:
```python
params = ans
extra_attrs = []
attrs, vals = zip(*sequences.items())
attr_per_param = [{k: val for k, val in zip(attrs, vs)} for vs in zip(*vals)]
if len(attr_per_param) != len(params):
    raise
for param, attrs in zip(params, attr_per_param):
    for name, val in attrs.items():
        setattr(param, name, val)
```
Look up how zip deals with too short sequences though. Now you could maybe end up with some params getting an e.g. `max` but not all.

symfit/core/minimizers.py
tBuLi: I kinda disagree, because I want to make sure people find enough examples of this syntax in the docs. But you are right, this docstring does not even explain how to use it directly which would make more sense.

symfit/core/minimizers.py
tBuLi: good point

symfit/core/support.py
tBuLi: agree

symfit/core/support.py
tBuLi: I think they are the most readable though, and we don't need min and max built-ins in this context. So it's something I personally can live with

symfit/core/support.py
tBuLi: Zip stops as soon as one of them terminates, so some explicit checking is required. However, I think I have a nice solution, soon to be pushed

symfit/core/minimizers.py
tBuLi: I will add a lot more comments. Not sure if there is really a different way though to achieve this result.

symfit/core/minimizers.py
pckroon: I'd argue that the docstring should describe on how to use this function/class. The actual docs should use the *best* way of using it.

symfit/core/support.py
pckroon: As it is now it's better, but still not there yet I think:
```python
sequences = {'min': iter([1, 2]), 'max': [1, 2, 3]}
names = ['a', 'b', 'c']
```

Also, make an if/else instead of assert. 

```python
params = ans
extra_attrs = []
attrs, vals = zip(*sequences.items())
attr_per_param = [{k: val for k, val in zip(attrs, vs)} for vs in zip(*vals)]
if len(attr_per_param) != len(params):
    raise
for param, attrs in zip(params, attr_per_param):
    if attrs.keys() != sequences.keys():
        raise
    for name, val in attrs.items():
        setattr(param, name, val)
```

tests/test_support.py
```diff
@@ -186,6 +188,31 @@ def test_parameters(self):
...
+        del a.f
+        # check that deletion was successful
+        with self.assertRaises(AttributeError):
+            # Does not exist before f is called
+            a._f
+        # However, the function should still be there
+        self.assertEqual(a.f, 2)
+        with self.assertRaises(AttributeError):
+            # Setting is not allowed.
+            a.f = 3
```
pckroon: Add a test in which A.f increments a counter so we can make sure the fucntion is not called unnecessarily.

tBuLi:
The performance problems have now been solved, by the introduction of a new decorator: `@cached_property`. The need for a double decorator 
```python
@property
@cache
def f():
    pass
```
always bothered me anyway, and this performance problem gave me a serious reason to look into it. The tests for this PR now run about a minute or more faster than the original ones, which didn't even use the expensive basinhopping yet.

The introduction of this decorator also solves the problem of how to clean the cache if a user changes `model_dict` after initial creation: we could make `model_dict` into a property and upon setting it, we simply call `del` on the relevant `cached_properties`.

tests/test_support.py
tBuLi: Good one

symfit/core/minimizers.py
tBuLi: changed to `method_name`, good point that the original name was not great.

docs/fitting_types.rst
```diff
@@ -416,35 +423,96 @@ your parameters. This can be incredibly annoying if you have no further
...
+
+No `minimizer_kwargs` have to be provided, as :mod:`symfit` will automatically
+compute and provide the jacobian and select a minimizer. In this case, `symfit`
+will choose `BFGS`. When bounds are provided, `symfit` will switch to
+using `L-BFGS-B` instead. Setting bounds is as simple as::
+
+    x1.min = 0.0
+    x1.max = 100.0
+
+However, the real strength the `symfit` syntax lies in providing constraints::
```
tBuLi: *of* the

symfit/core/fit.py
```diff
@@ -1333,21 +1316,28 @@ def __init__(self, model, *ordered_data, **named_data):
...
-            elif any([bound is not None for pair in self.model.bounds for bound in pair]):
-                # If any bound is set
-                minimizer = LBFGSB
-            else:
-                minimizer = BFGS
+            minimizer = self._determine_minimizer()
 
-        # Initialise the minimizer
-        if isinstance(minimizer, Sequence):
+        if isinstance(minimizer, Sequence):  # Initialise the minimizer
```
tBuLi: This comment should stay where it was

symfit/core/fit.py
```diff
@@ -1776,13 +1771,11 @@ def __neg__(self):
             new_model_dict[key] *= -1
         return self.__class__(new_model_dict, initial=self.initial)
 
-    @property
-    @cache
+    @cached_property
     def _ncomponents(self):
```
tBuLi: These guys really need docstrings

symfit/core/minimizers.py
```diff
@@ -429,30 +437,125 @@ class COBYLA(ScipyConstrainedMinimize):
...
+        
+        :param minimize_options: options to be passed on to
+            :func:`scipy.optimize.basinhopping`.
+        :return: :class:`symfit.core.fit_results.FitResults`
+        """
+        if 'minimizer_kwargs' not in minimize_options:
+            minimize_options['minimizer_kwargs'] = {}
+
+        if 'method' not in minimize_options['minimizer_kwargs']:
+            # If no minimizer was set by the user upon execute, use the default.
```
tBuLi: comment is wrong, we use the one provided as local_minimizer

tests/test_support.py
tBuLi: Present below this conversation

docs/fitting_types.rst
```diff
@@ -363,11 +364,12 @@ More common examples, such as dampened harmonic oscillators also work as expecte
...
 .. _global-fitting:
 
-Global Fitting
---------------
-In a global fitting problem, we fit to multiple datasets where one or more
-parameters might be shared. The same syntax used for ODE fitting makes this
-problem very easy to solve in :mod:`symfit`.
+Fitting multiple datasets
+-------------------------
+A common fitting problem, is to fit to multiple datasets. This is sometimes
```
pckroon: extraneous comma

docs/fitting_types.rst
```diff
@@ -397,6 +399,10 @@ normal way::
    :width: 500px
    :alt: ODE integration
 
+Any ``Model`` that comes to mind is fair game, behind the scenes :mod:`symfit`
```
pckroon: Comma should be period.

docs/fitting_types.rst
```diff
@@ -416,35 +422,96 @@ your parameters. This can be incredibly annoying if you have no further
...
+        df = np.zeros(2)
+        df[0] = -14.5 * np.sin(14.5 * x[0] - 0.3) + 2. * x[0] + 0.2
+        df[1] = 2. * x[1] + 0.2
+        return f, df
+
+    minimizer_kwargs = {"method":"L-BFGS-B", "jac":True}
+    x0 = [1.0, 1.0]
+    ret = basinhopping(func2d, x0, minimizer_kwargs=minimizer_kwargs, niter=200)
+
+Let's compare to the same functionality in :mod:`symfit`
```
pckroon: Missing `::`?

symfit/core/fit.py
pckroon: Jup

PR #180, opened 2018-09-29
Add Examples to docs

tBuLi:
By now there are many nice examples of `symfit` out there, mostly questions I answered on stackoverflow, and it is time to bundle them in a new `examples` directory in the docs so users can play around with them.

This is an non-exhaustive list of code snippets I know to be out there which could be turned into example scripts:

- [x] [Fourier Series](https://stackoverflow.com/questions/52524919/fourier-series-fit-in-python/52555801#52555801)
- [x] [Piecewise function](https://stackoverflow.com/questions/52030866/fitting-of-experimental-data-within-two-different-regions/52035987#52035987)
- [x] Polynomial Function
- [x] ODE Fitting
- [x] Interactive Guess examples


examples/piecewise.py
```diff
@@ -0,0 +1,34 @@
...
+model = Model({y: Piecewise((y1, t <= t0), (y2, t > t0))})
+
+# As a constraint, we demand equality between the two models at the point t0
+# to do this, we substitute t -> t0 and demand equality using `Eq`
+constraints = [Eq(y1.diff(t).subs({t: t0}), y2.diff(t).subs({t: t0}))]
+
+# # Generate example data
+tdata = np.linspace(0, 4., 200)
+ydata = model(t=tdata, a=63, b=300, d=2205, k=3, t0=0.65).y
+ydata = np.random.normal(ydata, 0.05 * ydata)  # add 1% noise
```
pckroon: 5% in the code

PR #182, opened 2018-10-05
Added CallableNumericalModel

tBuLi:
Added CallableNumericalModel, which is similar to CallableModel but whose dict components are python callables rather than sympy Expr.

Example:
```python
x, y = variables('x, y')
a, b = parameters('a, b')
numerical_model = CallableNumericalModel({y: lambda x, a, b: a * x + b}, independent_vars=[x], params=[a, b])
```
The great thing about this PR is that it allows powerusers to make models whose components have very radical possibilities, while still interfacing with `symfit` API in the same way.

In order to make this possible the inheritance structure behind models had to be reorganized somewhat, though not in a way which breaks backwards compatibility.


Jhsmit:
This PR is everything I've ever wanted. However, I can't get it to be work as I would want it yet. Short summary: my python callable returns a 2d matrix, my data is a 2d matrix of the same shape. I would expect symfit to be able to calculate a chi sq from this. However I get an error. When I flatten the input/output data, no error occurs. In principle its not a problem to flatten the arrays but AFAIK symfit should be shape agnostic.

Here is a MNWE:

```python

from symfit import Fit, Parameter, Variable
from symfit.core.fit import CallableNumericalModel
import numpy as np

shape = (30, 30)

def function(**parameters):
        out = np.ones(shape)*parameters['a']
        out[15:, :] += parameters['b']
        return out

a = Parameter(name='a', value=10)
b = Parameter(name='b', value=5)
y = Variable(name='y')

model = CallableNumericalModel({y: function}, [], [a, b])
data = 15*np.ones(shape)
data[15:, :] += 20

fit = Fit(model, y=data
res = fit.execute()
print(res)
```

tBuLi:
@Jhsmit, thanks for the observation, there was indeed a small bug in finite differences which assumed 1D data. Solved now, happy hunting.

Jhsmit:
It seems that when using BFGS minimizer bounds are ignored. In the previous example I've changed two lines:

```python
a = Parameter(name='a', value=10, max=12)
fit = Fit(model, y=data, minimizer=BFGS)
```

However, the result is 15.

tBuLi:
@pckroon: Those are good questions. I'll adress them one by one:

> Could I also make a mixed model where part is sympy-like, and part is numerical?

Come to think of it, yes, this already works because sympy expressions are callable in symfit. (We monky-patch `__call__`.) However, I will change `numerical_components` to reflect this because now there is no cashing of the lambdifying of such components.

> And maybe also interesting: What happens if the arguments to my lambda/function have different names than the variables/parameters I made? If these must be the same, is there a way around having to specify my variables and parameters to the call to CallableNumericalModel? 

The arguments on the callable and the independent_vars + params must have the same names, because symfit calls the model using keyword-arguments. And I also think it is the most readable. 

To answer the second part, no, there is no way around having to provide Parameter and Variable objects corresponding to the arguments of the function, because without using inspect there is no way for symfit to infer the names of the arguments. And I've learned my lesson about using inspect in production code ;).

> Lastly, what happens if I do something like this?
```python
def f(x, a, b=1):
    ...

x, y = variables()
a, c = parameters()
CallableNumericalModel({y: f}, [x], [a, c])
```
This will not run, and throws a `TypeError: <lambda>() got an unexpected keyword argument 'c'` or something similar.

But I think it is very reasonable to demand that the functions signature have to match the variables and parameters you tell symfit about. If a user has an existing function with different names, they could always be wrapped using a lambda. For example, in your example above:
```python
CallableNumericalModel({y: lambda x, a, c: f(x=x, a=a, b=c)}, [x], [a, c])
```

I should also remark that since all callables must accept all the arguments, it could be weird to type a very long list, e.g.:
```
CallableNumericalModel(
    {z1: lambda x, y, a, b, c: a*x},
    {z2: lambda x, y, a, b, c: a + b + c + x + y}, 
    [x, y], [a, b, c])
```
but this could easily be solved by refrazing `z1` to `z1: lambda x, a, **_: a*x`.

Lastly, there is also the issue that a user might want to provide additional args and kwargs which should be the same for every iteration. But for this, Python gave its followers `partial` :).

pckroon:
Ok, makes sense, and looks good to me. However, despite your learned lessons there's still inspect code around. But I guess that's necessary to make model signatures.

tBuLi:
Yeah that's inspect in a different role. It is now only used to generate signatures, not to read them. I've learned my lesson when it comes to reading them ;). 

I was told to generate them by watching a talk from David Beazley on youtube, so I think that's ok :).

I'll add one more test and then I'll rebase and add an example to the new example section of the docs as well. That should make this feature rock solid.

PR #183, opened 2018-10-07
Make sure zenodo gets the authors right.

tBuLi:
@pckroon do you have an ORCID you would like me to add?


pckroon:
`0000-0001-9273-4850`. And put my name as `Peter C Kroon`, if you please ;)

PR #185, opened 2018-10-12
Pickle fit results & Models

tBuLi:
FitResults turned out not to loadable after pickle, because of a wrongful implementation of __getattr__.

Additionally, models where no longer pickalable because of the cached lambda functions. This adresses these issues, and tests have been added.

### DIFF
--- a/symfit/core/fit.py
+++ b/symfit/core/fit.py
@@ -220,6 +220,9 @@ class BaseModel(Mapping):
                 del state[key]
         return state
 
+    def __reduce__(self):
+        return (self.__class__, (self.model_dict,))
+
 
 class BaseNumericalModel(BaseModel):
     """
@@ -273,6 +276,12 @@ class BaseNumericalModel(BaseModel):
     def shared_parameters(self):
         raise NotImplementedError(
             'Shared parameters can not be inferred for {}'.format(self.__class__.__name__)
+        )
+
+    def __reduce__(self):
+        return (
+            self.__class__,
+            (self.model_dict, self.independent_vars, self.params)
         )
 
 
@@ -742,6 +751,11 @@ class Constraint(Model):
         ]
         return inspect_sig.Signature(parameters=parameters)
 
+    def __reduce__(self):
+        return (
+            self.__class__,
+            (self.constraint_type(list(self.values())[0]), self.model)
+        )
 
 class TakesData(object):
     """
@@ -1828,6 +1842,17 @@ class ODEModel(CallableModel):
         self.sigmas = {var: Variable(name='sigma_{}'.format(var.name)) for var in self.dependent_vars}
 
         self.__signature__ = self._make_signature()
+
+    def __reduce__(self):
+        if self.lsoda_kwargs:
+            from pickle import PicklingError
+            raise PicklingError(
+                '{} with lsoda_kwargs can not be pickled.'.format(self.__class__)
+            )
+        return (
+            self.__class__,
+            tuple([self.model_dict, self.initial] + list(self.lsoda_args))
+        )
 
     def __str__(self):
         """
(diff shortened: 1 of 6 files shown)


symfit/core/fit_results.py
```diff
@@ -69,10 +69,10 @@ def __getattr__(self, item):
         :param item: Name of Goodness of Fit qualifier.
         :return: Goodness of Fit qualifier if present.
         """
-        if item in self.gof_qualifiers:
-            return self.gof_qualifiers[item]
-        else:
-            raise AttributeError
+        if 'gof_qualifiers' in vars(self):
```
pckroon: Why not just do:
`return self.gof_qualifiers[item]`
?
If the attribute is not there, AttributeError is raised anyway.

tBuLi: Upon pickle.load, gof_qualifiers does not yet exist and then this will cause an infinite recursion error. This is apparently the correct way to bypass it, see e.g. https://stackoverflow.com/questions/49380224/how-to-make-classes-with-getattr-pickable

tBuLi: The first suggestion will not work and give the same recursion error. The problem is with `got_qualifiers` not existing, not with the key not existing. You are right, if an attribute doesn't exist, `__getattr__` is called. Then, if that raises an `AttributeError` too, that's the end of it.

However, calling `self. gof_qualifiers` if `self. gof_qualifiers` does not exist, raises an AttributeError, which takes you to `__getattr__`. Then, in `__getattr__`, we look for `self. gof_qualifiers`. The result is infinite recursion.

`__getinitargs__` is only for old-style classes, which we don't use. It seems the py3 equivalent is `__reduce__`. Although that might also be worth adding so that `__init__` is called, I think this fix of `__getattr__` is also in order.

pckroon: I made a small testclass, and it seems you're right :'( (and it's slowly starting to make sense).
Creating a simple `__getstate__` and `__setstate__` also fixes it though. This at least keeps `pickle` specifics in the place you expect them.

```python
class A:
    def __init__(self):
        self.attr = {}
        self.attr['a'] = 1

    def __getattr__(self, item):
        try:
            return self.attr[item]
        except KeyError as error:
            raise AttributeError from error

    def __getstate__(self):
        return self.__dict__.copy()

    def __setstate__(self, state):
        self.__dict__.update(state)


a = A()

p = pickle.dumps(a)

a2 = pickle.loads(p)
```

tBuLi: I agree that maybe that should be done too, but updating `__getattr__` in this way does make sure this function is at least consistent, do you agree?

I think it is weird that such minimal getstate and setstate change anything btw, because this is kinda what I expect python to do anyway.

tBuLi: Implementing `__reduce__` on all models such that init is called upon pickle.load seems to solve all the problems. If you agree with this solution then I will merge this PR so our mystery power-user @Jhsmit will be happy.

I will then also make a new release, 0.4.5.1, to trigger a new Zenodo DOI build to see if that has been resolved.

symfit/core/fit.py
```diff
@@ -742,6 +751,11 @@ def _make_signature(self):
         ]
         return inspect_sig.Signature(parameters=parameters)
 
+    def __reduce__(self):
+        return (
+            self.__class__,
+            (self.constraint_type(list(self.values())[0]), self.model)
+        )
```
pckroon: Should this also call `super().__reduce__`? Same goes for all the other reduce methods

tBuLi: It could, but I think it would make this way less readable since tuples are immutable and this has to return a tuple of tuples.

pckroon: Ok, do you then actually manage to fully reproduce the object?

PR #189, opened 2018-10-19
Implement trust-constr wrapper

pckroon:
This implements the trust-constr method from scipy. They advertise it as the most robust/flexible/powerfull minimizer in their arsenal. It's kind of tricky to make it play ball, but I think this should work in most cases.

### DIFF
--- a/symfit/core/fit.py
+++ b/symfit/core/fit.py
@@ -1500,9 +1500,9 @@ class Fit(HasCovarianceMatrix):
             if hasattr(self.model, 'numerical_jacobian') and hasattr(self.objective, 'eval_jacobian'):
                 minimizer_options['jacobian'] = self.objective.eval_jacobian
         if issubclass(minimizer, HessianMinimizer):
-            # If an analytical version of the Jacobian exists we should use
+            # If an analytical version of the Hessian exists we should use
             # that, otherwise we let the minimizer estimate it itself.
-            # Hence the check of numerical_jacobian, as this is the
+            # Hence the check of numerical_hessian, as this is the
             # py function version of the analytical jacobian.
             if hasattr(self.model, 'numerical_hessian') and hasattr(self.objective, 'eval_hessian'):
                 minimizer_options['hessian'] = self.objective.eval_hessian
--- a/symfit/core/fit.py
+++ b/symfit/core/fit.py
@@ -15,8 +15,8 @@ from .support import (
     seperate_symbols, keywordonly, sympy_to_py, key2str, partial, cached_property
 )
 from .minimizers import (
-    BFGS, SLSQP, LBFGSB, BaseMinimizer, GradientMinimizer, ConstrainedMinimizer,
-    ScipyMinimize, MINPACK, ChainedMinimizer, BasinHopping
+    BFGS, SLSQP, LBFGSB, BaseMinimizer, GradientMinimizer, HessianMinimizer,
+    ConstrainedMinimizer, MINPACK, ChainedMinimizer, BasinHopping
 )
 from .objectives import (
     LeastSquares, BaseObjective, MinimizeModel, VectorLeastSquares, LogLikelihood
@@ -1499,6 +1499,13 @@ class Fit(HasCovarianceMatrix):
             # py function version of the analytical jacobian.
             if hasattr(self.model, 'numerical_jacobian') and hasattr(self.objective, 'eval_jacobian'):
                 minimizer_options['jacobian'] = self.objective.eval_jacobian
+        if issubclass(minimizer, HessianMinimizer):
+            # If an analytical version of the Jacobian exists we should use
+            # that, otherwise we let the minimizer estimate it itself.
+            # Hence the check of numerical_jacobian, as this is the
+            # py function version of the analytical jacobian.
+            if hasattr(self.model, 'numerical_hessian') and hasattr(self.objective, 'eval_hessian'):
+                minimizer_options['hessian'] = self.objective.eval_hessian
 
         if issubclass(minimizer, ConstrainedMinimizer):
             # set the constraints as MinimizeModel. The dependent vars of the
--- a/symfit/core/minimizers.py
+++ b/symfit/core/minimizers.py
@@ -132,7 +132,7 @@ class GradientMinimizer(BaseMinimizer):
             self.jacobian = self._baseobjective_from_callable(self.jacobian)
             self.wrapped_jacobian = self.resize_jac(self.jacobian)
         else:
-            self.jacobian = None
+#            self.jacobian = None
             self.wrapped_jacobian = None
 
     def resize_jac(self, func):
@@ -143,6 +143,38 @@ class GradientMinimizer(BaseMinimizer):
         :param func: Jacobian function to be wrapped. Is assumed to be the
             jacobian of a scalar function.
         :return: Jacobian corresponding to non-fixed parameters only.
+        """
+        if func is None:
+            return None
+        @wraps(func)
+        def resized(*args, **kwargs):
+            out = func(*args, **kwargs)
+            # Make one dimensional, corresponding to a scalar function.
+            out = np.atleast_1d(np.squeeze(out))
+            mask = [p not in self._fixed_params for p in self.parameters]
+            return out[mask]
+        return resized
+
+
+class HessianMinimizer(GradientMinimizer):
+    @keywordonly(hessian=None)
+    def __init__(self, *args, **kwargs):
+        hessian = kwargs.pop('hessian')
+        super(HessianMinimizer, self).__init__(*args, **kwargs)
+        if hessian is not None:
+            hess_with_fixed_parameters = partial(hessian, **{p.name: p.value for p in self._fixed_params})
+            self.wrapped_hessian = self.resize_hess(hess_with_fixed_parameters)
+        else:
+            self.wrapped_hessian = None
+
+    def resize_hess(self, func):
+        """
+        Removes values with identical indices to fixed parameters from the
+        output of func. func has to return the Hessian of a scalar function.
+
+        :param func: Hessian function to be wrapped. Is assumed to be the
+            Hessian of a scalar function.
+        :return: Hessian corresponding to non-fixed parameters only.
         """
         if func is None:
             return None
@@ -276,13 +308,10 @@ class ScipyMinimize(object):
     Mix-in class that handles the execute calls to :func:`scipy.optimize.minimize`.
     """
     def __init__(self, *args, **kwargs):
-        self.constraints = []
-        self.jacobian = None
-        self.wrapped_jacobian = None
         super(ScipyMinimize, self).__init__(*args, **kwargs)
 
     @keywordonly(tol=1e-9)
-    def execute(self, bounds=None, jacobian=None, constraints=None, **minimize_options):
+    def execute(self, bounds=None, jacobian=None, hessian=None, constraints=None, **minimize_options):
         """
         Calls the wrapped algorithm.
 
@@ -301,6 +330,7 @@ class ScipyMinimize(object):
             bounds=bounds,
             constraints=constraints,
             jac=jacobian,
+            hess=hessian,
             **minimize_options
         )
         return self._pack_output(ans)
@@ -361,9 +391,61 @@ class ScipyGradientMinimize(ScipyMinimize, GradientMinimizer):
     """
     def __init__(self, *args, **kwargs):
         super(ScipyGradientMinimize, self).__init__(*args, **kwargs)
+<<<<<<< HEAD
+=======
+        self.has_jacobian = self.wrapped_jacobian is not None
+        self.wrapped_jacobian = self.list2kwargs(self.wrapped_jacobian)
+>>>>>>> 821a77a... Take a broom through minimizer inheritence
 
+    @keywordonly(jacobian=None)
     def execute(self, **minimize_options):
-        return super(ScipyGradientMinimize, self).execute(jacobian=self.wrapped_jacobian, **minimize_options)
+        jacobian = minimize_options.pop('jacobian')
+        if jacobian is None:
+            jacobian = self.wrapped_jacobian
+        return super(ScipyGradientMinimize, self).execute(jacobian=jacobian, **minimize_options)
+
+class ScipyHessianMinimize(ScipyGradientMinimize, HessianMinimizer):
+    """
+    Base class for :func:`scipy.optimize.minimize`'s hessian-minimizers.
+    """
+    def __init__(self, *args, **kwargs):
+        super(ScipyHessianMinimize, self).__init__(*args, **kwargs)
+        self.has_hessian = self.wrapped_hessian is not None
+        self.wrapped_hessian= self.list2kwargs(self.wrapped_hessian)
+    
+    @keywordonly(jacobian=None, hessian=None)
+    def execute(self, **minimize_options):
+        hessian = minimize_options.pop('hessian')
+        jacobian = minimize_options.pop('jacobian')
+        options = minimize_options.get('options', {})
+
+        # Our Jacobians are dense, and apparently we need to explicitely
+        # tell this.
+        if self.has_jacobian:
+            options.update(sparse_jacobian=False)
+
+        # For models that are not differentiable, users need the ability to
+        # change the jacobian to e.g. 'cs' or '3-point'. In that case, hess
+        # should either be scipy.optimize.BFGS or SR1.
+        # In addition, users may want to change the way the Hessian is
+        # calculated, especially if they manage to make a model whose Jacobian
+        # can't handle complex numbers.
+        if self.has_jacobian and not self.has_hessian:
+            if hessian is None:
+                hessian = 'cs'
+        elif not self.has_jacobian and not self.has_hessian:
+            if jacobian is None:
+                jacobian = 'cs'
+            if hessian is None:
+                hessian = so.BFGS(exception_strategy='damp_update')
+
+        if jacobian is None:
+            jacobian = self.wrapped_jacobian
+        if hessian is None:
+            hessian = self.wrapped_hessian
+
+        return super(ScipyHessianMinimize, self).execute(jacobian=jacobian, hessian=hessian, **minimize_options)
+
 
 class ScipyConstrainedMinimize(ScipyMinimize, ConstrainedMinimizer):
     """
@@ -391,6 +473,7 @@ class ScipyConstrainedMinimize(ScipyMinimize, ConstrainedMinimizer):
             sympy.Eq: 'eq', sympy.Ge: 'ineq',
         }
 
+<<<<<<< HEAD
         for constraint in constraints:
             if isinstance(constraint, MinimizeModel):
                 # Typically the case when called by `Fit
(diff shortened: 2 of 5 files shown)


pckroon:
Also, if this works as intended/promised, we can consider simplifying Fit by making it always use trust-constr. 
Lastly, it may be worthwhile to also generate the Hessian analytically.

symfit/core/minimizers.py
```diff
@@ -465,7 +466,76 @@ class Powell(ScipyMinimize, BaseMinimizer):
...
+    Wrapper around :func:`scipy.optimize.minimize`'s Trust-Constr algorithm.
+    """
+    def __init__(self, *args, **kwargs):
+        super(TrustConstr, self).__init__(*args, **kwargs)
+        # We have to break DRY because you cannot inherit from both
+        # ScipyConstrainedMinimize and ScipyGradientMinimize. So SLQSP is a
+        # special case. This is the same code as in ScipyGradientMinimize.
+        self.wrapped_jacobian = self.list2kwargs(self.wrapped_jacobian)
+
+    @keywordonly(jacobian=None, hess=None, options=None)
```
tBuLi: I'd use `hessian` in full, in keeping with using `jacobian` instead of `jac` like scipy does.

symfit/core/minimizers.py
```diff
@@ -465,7 +466,76 @@ class Powell(ScipyMinimize, BaseMinimizer):
...
+    def execute(self, **minimize_options):
+        jacobian = minimize_options.pop('jacobian')
+        hess = minimize_options.pop('hess')
+        options = minimize_options.pop('options')
+        if options is None:
+            options = {}
+        # For models that are not differentiable, users need the ability to
```
tBuLi: Good idea, but I think the logic pertaining to the jacobian should be moved to `ScipyGradientMinimize` so it can be used for any gradient method?

symfit/core/minimizers.py
```diff
@@ -382,79 +461,40 @@ def scipy_constraints(self, constraints):
...
-            scipy_constraint['jac'] = lambda p, c: self.resize_jac(
-                    self.list2kwargs(
-                        partial(c.func.eval_jacobian, **partialed_kwargs)
-                    )
-                )(list(p))
-        return scipy_constr
-
-
-class COBYLA(ScipyConstrainedMinimize):
+class COBYLA(ScipyConstrainedMinimize, BoundedMinimizer):
```
tBuLi: `COBYLA` does not support bounds according to the docs. https://docs.scipy.org/doc/scipy/reference/optimize.minimize-cobyla.html#optimize-minimize-cobyla

tBuLi:
In light of #195, I think it is best that I first complete and merge that one, and then I can experiment with doing a rebase of this PR on that new master.

As a general question, a lot of the methods now inherit from `BoundedMinimizer`, even though they do not support bounds according to the scipy docs. Why is that?

symfit/core/minimizers.py
```diff
@@ -129,6 +130,38 @@ def resized(*args, **kwargs):
...
+    def __init__(self, *args, **kwargs):
+        hessian = kwargs.pop('hessian')
+        super(HessianMinimizer, self).__init__(*args, **kwargs)
+        if hessian is not None:
+            hess_with_fixed_parameters = partial(hessian, **{p.name: p.value for p in self._fixed_params})
+            self.wrapped_hessian = self.resize_hess(hess_with_fixed_parameters)
+        else:
+            self.wrapped_hessian = None
+
+    def resize_hess(self, func):
```
tBuLi: I think this code is identical to resize_jac, correct? Since we inherit from `GradientMinimizer`, why not just call resize_jac? After the resize has happened nobody will ever know anyway, since the new function will mascarade as `resized`.

symfit/core/minimizers.py
```diff
@@ -3,6 +3,7 @@
 from collections import namedtuple, Counter
 
 from scipy.optimize import minimize, differential_evolution, basinhopping
+import scipy.optimize as so
```
tBuLi: why not import in the style as the line above this one? Or, if that is starting to lead to namespace conflict we should rename everything to `so.something` (which I would prefer not to do)

tests/test_constrained.py
```diff
@@ -610,5 +610,74 @@ def test_fixed_and_constrained(self):
             self.assertEqual(obj_jac.shape, cons_jac.shape)
             self.assertEqual(obj_jac.shape, (2,))
 
+    def test_fixed_and_constrained_tc(self):
```
tBuLi: Is this test in anyway different from `test_fixed_and_constrained`, or is it identical? Cause if so, I think it would be better to add a small for loop to `test_fixed_and_constrained` to run over both minimizers.

symfit/core/minimizers.py
pckroon: I need access to scipy.optimize.BFGS, which conflicts with our BFGS. I'll look for an alternative (`from so import BFGS as soBFGS`)

tests/test_constrained.py
pckroon: Similar but different, because trust-constr constraints are entirely different from COBYLA/SLSSQP constraints. Literally 0 overlap.

pckroon:
I was under the impression that all scipy minimizers could deal with bounds. But apparently not. Note that this does not lead to failing tests...

symfit/core/minimizers.py
```diff
@@ -373,88 +435,62 @@ def scipy_constraints(self, constraints):
...
-                # evaluates the constraint with these parameters.
-                # Wrap `c` so it is always called via keywords.
-                'fun': lambda p, c: self.list2kwargs(c)(list(p))[0],
-                'args': [partialed_constraint]
-            })
-
+                'fun': constraint,
+                }
+            if isinstance(self, ScipyGradientMinimize):
+                con['jac'] = self.resize_jac(constraint.eval_jacobian)
```
pckroon: I'm not sure this is the right place for this. Maybe move it to ScipyGradientMinize (and ScipyHessianMinimize)?

symfit/core/minimizers.py
```diff
@@ -129,6 +157,20 @@ def resized(*args, **kwargs):
         return resized
 
 
+class HessianMinimizer(GradientMinimizer):
```
pckroon: Missing Docstring

symfit/core/minimizers.py
```diff
@@ -358,11 +412,86 @@ class ScipyGradientMinimize(ScipyMinimize, GradientMinimizer):
...
-        super(ScipyGradientMinimize, self).__init__(*args, **kwargs)
+    @keywordonly(jacobian=None)
+    def execute(self, **minimize_options):
+        if 'jacobian' in minimize_options:
+            jacobian = minimize_options.pop('jacobian')
+        else:
+            jacobian = self.wrapped_jacobian
+        return super(ScipyGradientMinimize, self).execute(jacobian=jacobian, **minimize_options)
+
+    def scipy_constraints(self, constraints):
```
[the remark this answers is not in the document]
pckroon: It deals with the gradient of the constraints. As such I thought it was better with ScipyGradientMinimize than with ScipyConstraintMinimize. The same goes for the Hessian minimizer actually, but there there's a bit more logic to choose the method to calculate the jac/hess

symfit/core/minimizers.py
```diff
@@ -358,11 +412,86 @@ class ScipyGradientMinimize(ScipyMinimize, GradientMinimizer):
...
     """
-    def __init__(self, *args, **kwargs):
-        super(ScipyGradientMinimize, self).__init__(*args, **kwargs)
+    @keywordonly(jacobian=None)
+    def execute(self, **minimize_options):
+        if 'jacobian' in minimize_options:
+            jacobian = minimize_options.pop('jacobian')
+        else:
+            jacobian = self.wrapped_jacobian
+        return super(ScipyGradientMinimize, self).execute(jacobian=jacobian, **minimize_options)
```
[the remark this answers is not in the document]
pckroon: For the ScipyHessianMinimizer the user needs to be able to override the jacobian/hessian calculation.

symfit/core/minimizers.py
```diff
@@ -358,11 +412,86 @@ class ScipyGradientMinimize(ScipyMinimize, GradientMinimizer):
...
+            hessian = 'cs'
+        elif not self.has_jacobian and not self.has_hessian:
+            jacobian = 'cs'
+            hessian = soBFGS(exception_strategy='damp_update')
+        else:
+            jacobian = None
+            hessian = None
+        return jacobian, hessian
 
+    @keywordonly(jacobian=None, hessian=None)
```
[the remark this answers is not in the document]
pckroon: For the ScipyHessianMinimizer the user needs to be able to override the jacobian/hessian calculation, although mostly just for the trust-constr minimizer. Maybe I'll move it there.

symfit/core/minimizers.py
```diff
@@ -358,11 +412,86 @@ class ScipyGradientMinimize(ScipyMinimize, GradientMinimizer):
...
+        repectively. If None, it should be calculated using the available
+        analytical method.
+
+        :return: tuple of jacobian_method, hessian_method
+        """
+        if self.has_jacobian and not self.has_hessian:
+            hessian = 'cs'
+        elif not self.has_jacobian and not self.has_hessian:
+            jacobian = 'cs'
+            hessian = soBFGS(exception_strategy='damp_update')
```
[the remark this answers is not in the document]
pckroon: There's three cases:
1) Both the Jacobian and Hessian are provided: Do nothing special
2) Only the Jacobian is provided: approximate the Hessian using a complex step finite difference ('cs')
3) Neither is provided: approximate the Jacobian using a complex step finite difference ('cs'). In that case scipy doesn't let us approximate the Hessian using finite difference, and instead forces you to use a numerical optimization. There are 2 minimizers that can do this, `scipy.optimize.BFGS` and `SR1`. SR1 blew up in my face so I abandoned that one. `so.BFGS` requires an option to be passed for `exception_strategy`.

This is also the reason why execute must accept a `jacobian` and `hessian` kwarg for the user to override this behaviour, e.g. when their function can not be evaluated in the complex plane.

Note: all this is only valid for the `trust-constr` minimizer. So maybe I'll move it there.

symfit/core/minimizers.py
```diff
@@ -414,60 +543,28 @@ def scipy_constraints(self, constraints):
...
-            })
-
+                }
+            cons.append(con)
         cons = tuple(cons)
         return cons
 
 
-class BFGS(ScipyGradientMinimize):
+class BFGS(ScipyGradientMinimize, GradientMinimizer):
```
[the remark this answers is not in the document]
pckroon: See different comment (to be written)

symfit/core/minimizers.py
```diff
@@ -494,14 +585,56 @@ class NelderMead(ScipyMinimize, BaseMinimizer):
...
+
+    def scipy_constraints(self, constraints):
+        cons = super(TrustConstr, self).scipy_constraints(constraints)
+        out = []
+        for con in cons:
+            if con['type'] == 'eq':
+                ub = 0
+            else:
+                ub = np.inf
+            tc_con = NonlinearConstraint(fun=con['fun'],
```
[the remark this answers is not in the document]
pckroon: It's a journey of discovery and surprise every single time.

symfit/core/minimizers.py
```diff
@@ -494,14 +585,56 @@ class NelderMead(ScipyMinimize, BaseMinimizer):
...
+
+    def execute(self, **minimize_options):
+        options = minimize_options.get('options', {})
+        # Our Jacobians are dense, and apparently we need to explicitely
+        # tell this.
+        options.update(sparse_jacobian=False)
+        minimize_options['options'] = options
+        return super(TrustConstr, self).execute(**minimize_options)
+
+class DifferentialEvolution(ScipyMinimize, GlobalMinimizer, BoundedMinimizer):
```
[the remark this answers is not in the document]
pckroon: Should be

tests/test_minimizers.py
```diff
@@ -139,7 +156,9 @@ def test_pickle(self):
         constrained_minimizers = subclasses(ScipyConstrainedMinimize)
         # Test for all of them if they can be pickled.
         for minimizer in scipy_minimizers:
-            if minimizer in constrained_minimizers:
+            # TrustConstr's constraints are wildly different
+            # TODO: Actually fix the tests in that case
```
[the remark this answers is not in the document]
pckroon: Just different. The constraints are so different that these tests can't deal with it.

tests/test_general.py
```diff
@@ -617,7 +617,7 @@ def test_likelihood_fitting_gaussian(self):
 
         self.assertAlmostEqual(fit_result.value(mu) / mean, 1, 6)
         self.assertAlmostEqual(fit_result.stdev(mu) / mean_stdev, 1, 3)
-        self.assertAlmostEqual(fit_result.value(sig) / np.std(xdata), 1, 6)
+        self.assertAlmostEqual(fit_result.value(sig) / np.std(xdata), 1, 5)
```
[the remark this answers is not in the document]
pckroon: I have *no* idea what's happening here. I tested a few things compared to master, and I found that

1) The jacobian is rather sensitive to the parameter values
2) The jacobian produces the same values between both branches (master and trust-constr)
3) Both use the LBFGSB minimizer.
4) Bounds are the same, and *are* passed to the execute call

So I give up unless you have more ideas on what to look for.

symfit/core/minimizers.py
```diff
@@ -293,13 +332,28 @@ def execute(self, bounds=None, jacobian=None, constraints=None, **minimize_optio
             :func:`scipy.optimize.minimize`. Note that your `method` will
             usually be filled by a specific subclass.
         """
+        if bounds is None and isinstance(self, BoundedMinimizer):
```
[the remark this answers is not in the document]
pckroon: Either way, I think this is getting out of scope for this PR.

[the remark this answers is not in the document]
tBuLi: I think the idea of having one deal with init, and the other with execute, is a nice idea. I just don't know if it will scale, but for now that would work for all our objects.

Between those two options, I think option 1 is the cleanest. The inheritance could also be drastically cleanup by having methods inherit from each other, which is code technically the right thing to do although conceptually it is weird.

For example, 
```python
class BFGS(GradientMinimize, ScipyGradientMinimize):
    pass

class LBFGSB(BFGS, BoundedMinimize, ScipyBoundedMinimize):
    pass

class SLSQP(LBFGSB, ConstrainedMinimize, ScipyConstrainedMinimize):
    pass
```
Codewise this is a lot cleaner although it is not true that SLSQP inherits from LBFGSB as an algorithm.

Looking at this code now though, it does seem a bit silly to type things twice. Maybe we can make a seperate PR to dive into the MRO again, there has to be a better way.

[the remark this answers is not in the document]
pckroon: This is not something to rush into indeed. I'd say this is out of scope for this PR. We should let it stew for a while, and revisit the minimizer MRO/inheritance tree again later.
Let's leave this as is for now (it works, yay).

symfit/core/minimizers.py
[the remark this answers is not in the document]
tBuLi: I think this is part of a more general problem: when the user wants to provide a different value to an input which symfit already defines by default, we have the potential for conflict. I think this is a problem we should solve generally, but which to my mind does not need to be solved immediately because in symfit we can typically provide the exact Hessian from the model, and in other cases the scipy default will do just fine. Unless I'm missing why this particular algorithm would be way more sensitive to this.

symfit/core/minimizers.py
```diff
@@ -358,11 +398,46 @@ class ScipyGradientMinimize(ScipyMinimize, GradientMinimizer):
...
+        if hessian is None:
+            hessian = self.wrapped_hessian
+        return super(ScipyHessianMinimize, self).execute(hessian=hessian, **minimize_options)
+
+    def scipy_constraints(self, constraints):
+        cons = super(ScipyHessianMinimize, self).scipy_constraints(constraints)
+        # FIXME: Just assume that because the objective has a hess, so do
+        # all the constraints
+        for con in cons:
+            con['hess'] = self.resize_hess(con['fun'].eval_hessian)
```
tBuLi: I do not think such a key exists: https://docs.scipy.org/doc/scipy-1.1.0/reference/generated/scipy.optimize.minimize.html#scipy.optimize.minimize

symfit/core/minimizers.py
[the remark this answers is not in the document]
tBuLi: ok, that is another way to look at it. Not sure there is a right choice here actually, unless if we introduce a `ScipyConstrainedGradientMinimize`.

symfit/core/minimizers.py
```diff
@@ -494,14 +531,103 @@ class NelderMead(ScipyMinimize, BaseMinimizer):
...
+            hessian = soBFGS(exception_strategy='damp_update')
+        else:
+            jacobian = None
+            hessian = None
+        return jacobian, hessian
+
+    def scipy_constraints(self, constraints):
+        cons = super(TrustConstr, self).scipy_constraints(constraints)
+        # TODO: If no hessian/jac for this model, assume no hessian/jac for the
+        # constraints, and do things with self._get_jacobian_hessian_strategy
```
tBuLi: I'm not sure that is necessary actually, at least not for now. Constraints are `Model` subclasses so they have a jacobian and hessian.

However, if in the future any model could be given as a constraint, including numerical ones, then this does become an issue

tests/test_constrained.py
```diff
@@ -629,5 +629,86 @@ def test_fixed_and_constrained(self):
...
+        """
+        Taken from #165. Fixing parameters and constraining others caused a
+        TypeError: missing a required argument: 'theta1', which was caused by a
+        mismatch in the shape of the initial guesses given and the number of
+        parameters constraints expected. The initial_guesses no longer contained
+        those corresponding to fixed parameters.
+        """
```
tBuLi: perhaps update this docstring

symfit/core/minimizers.py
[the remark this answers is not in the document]
pckroon: In a nutshell: the scipy default are not sufficient. If you pass not jacobian/hessian information to trust-constr it will blow up into your face saying it needs both.

In my view you should be able to use any of our wrapped minimizers with any kind of model, even those for which we don't know a jacobian/hessian analytically. That means that in those cases you absolutely must use a gradient/hessian based minimizer you need to be able to provide a jacobian/hessian. In that case the responsibility is on the user to 1) make sure they actually do, and 2) whatever they pass conforms to the API and 3) produces the correct values though.

In any case, if the user thinks they know better than what we provide for them, we should honor that since they have more intimate knowledge/information about their system/model. It's up to them to analyze the ashes once it blows up though.

So to wrap up, this is not an issue in 95% of the cases where the sane symfit determined analytical jacobian/hessian are passed. Allowing users to override our defaults enables them to also use this is the other 5% of the cases though.

symfit/core/minimizers.py
pckroon: That is because currently the only Hessian method that can deal with constraints is trust-constr, which coincidentally uses a wildly different constraint format.
For the sake of consistency I think this should be in here, and it might catch some errors where a user tries to do a constraint hessian based minimization using something else than trust-constr.

symfit/core/minimizers.py
pckroon: I agree with that analysis, which is why I put a TODO instead of trying to figure out a solution.

symfit/core/fit.py
```diff
@@ -1499,6 +1499,13 @@ def _init_minimizer(self, minimizer, **minimizer_options):
             # py function version of the analytical jacobian.
             if hasattr(self.model, 'numerical_jacobian') and hasattr(self.objective, 'eval_jacobian'):
                 minimizer_options['jacobian'] = self.objective.eval_jacobian
+        if issubclass(minimizer, HessianMinimizer):
+            # If an analytical version of the Hessian exists we should use
+            # that, otherwise we let the minimizer estimate it itself.
+            # Hence the check of numerical_hessian, as this is the
+            # py function version of the analytical jacobian.
+            if hasattr(self.model, 'numerical_hessian') and hasattr(self.objective, 'eval_hessian'):
+                minimizer_options['hessian'] = self.objective.eval_hessian
```
pckroon: This contradicts the comment. And is numerical_hessian still a thing or did it get trimmed recently?

symfit/core/minimizers.py
```diff
@@ -155,6 +154,42 @@ def resized(*args, **kwargs):
         return resized
 
 
+class HessianMinimizer(GradientMinimizer):
+    """
+    ABC for Minimizer's that support the use of a Hessian.
```
pckroon: typo

symfit/core/fit.py
tBuLi: In #201 the check for `model.numerical_xxx` will be replaced by `model.eval_xxx`. But I think we should merge this PR first, and then I will rebase #201. In #201 I have removed `numerical_jacobian` and `numerical_hessian` in favor of giving Hessians and Jacobians there own model, so this well definitely make some tests fail so I cannot forget to update this line. So I would say leave it like this for this PR.

pckroon: > So I would say leave it like this for this PR.

Best possible answer :D

tBuLi:
Merged. One thing this PR made very obvious though, is that we need to take another look at the minimizer structure. I'll open a new issue where we can discuss this.

pckroon:
Completely agree. Please copy some of the discussion/ideas we've had on that here to that issue. Should save us some work.

pckroon:
Oh, and thanks for the help in polishing this :)

tBuLi:
You're welcome, implementing this was a good move.

PR #190, opened 2018-10-19
Add analytical Hessian to Model. Clean up Jacobian reshaping

pckroon:
The Hessian still needs to be added to the rest of the machinery: Objectives and Minimizers; but I want a hand with that. Especially for Objectives the maths is annoying


tBuLi:
Which objectives do you think require this? I think `LeastSquares` and `MinimizeModel` should do?

pckroon:
And maybe `LogLikelihood` as well. It depends on how much time you have available and how hard it is ;)

tBuLi:
Getting on it now

tBuLi:
Superseded and solved in #199.

Issue #194, opened 2018-10-28
Allow interdependence of components


tBuLi:
Consider the following matrix equation:
```python
from symfit import symbols, Parameter, MatrixSymbol, Inverse, Fit

N = symbols('N', integer=True)
M = MatrixSymbol('M', N, N)
I = MatrixSymbol('I', N, N)
y = MatrixSymbol('y', N, 1)
c = MatrixSymbol('c', N, 1)
a = Parameter('a')

model_dict = {c: - Inverse(I + M / a**2) * y}
```
The quantity I would like to fit is the scalar `z = c.T * c = d**2`, where `d` is some experimental value. Here is the issue: the value `c` contains a matrix inversion, which is expensive to compute, and in `z` this is performed twice. However, the usual solution of substituting `W = Inverse(I + M / a**2)` and calculating that separately is not available to us, because it is a function of the parameter `a`.

The most natural thing to do would be
```python
model_dict = {
    W: Inverse(I + M / a**2),
    c: - W * y,
    z: c.T * c
}
model = Model(model_dict)
fit = Fit(model_dict, z=d**2, I=..., M=..., y=...)
```
This will not require any changes to `Fit`, because we already allow this for `ODEModels`. There we also sometimes have dummy components with no data, and they are ignored. In the model above, since no data is provided for the components `c` and `W`, they too will be ignored by `Fit`.

The problems are that `Model`'s signature can not be generated, because of the double roles played by `c` and `W` as both dependent and independent variables, and when evaluating `__call__` the components will have to be calculated in the correct order and substituted.

Analogously to `ODEModel`, I think `c` and `W` should be considered as dependent variables.

Then, the lambda functions per component should somehow be linked, preferably without destroying backwards compatibility to code which is calling numerical_components directly instead of through `__call__`.

pckroon:
It might be worth to try an implementation that uses networkx, and if that works see how much functionality we'd need to reproduce.

tBuLi:
Those are some good points to think about, especially the cyclic dependencies. But I'm not sure that those systems can even be built, intuitively I think that such a system would be singular?

tBuLi:
After some more experimentation, I now have a working version for `CallableModel`'s. However, for `Model` all hell breaks loose because the jacobian can not be determined since sympy does not know about these dependencies! That means that the above example will have to be changed to something like:

```python
N = symbols('N', integer=True)
M = MatrixSymbol('M', N, N)
I = MatrixSymbol('I', N, N)
[...]

Issue #154, opened 2018-05-07
`Constraints` should not be a seperate type.


tBuLi:
`Constraint`'s are fundamentally just `Model`'s and so should not have a seperate role. They were introduced mostly because they should be derived w.r.t. all the parameters of the associated model, not of the constraint itself.

However, at the time of coding this I had forgotten that python is a consenting adults language. So we could just initiate a normal `Model` with the expression of the constraint, and then overwrite its `.params` and `.vars` with those of the associated model.

This does leave the issue of what to do about the two different fundamental types: `Eq` or `GreaterThan`. This could be either set as an attribute on the model, or we track the different types in the calling function (`Fit`).

pckroon:
The advantage of having separate `Constraints` is ease of use on the user side. So if you want to remove the class `Constraint` it should at least be replaced by a convenience function that does what you describe (something like: `add_constraint(model, constraint)`; or even make it a method of `Model`).

tBuLi:
Not sure that it would impact the user, as users already don't interact with `Constraint`'s directly. The API would not change, it would just be that the current list `Fit.constraints` list would now contain `Model`'s rather than `Constraint`'s. Although this list might now change have to into a dict to be able to separate the types of constraint.

I do intend to make a seperate convinience function which will take care of overwriting the .params etc, though I think it should take a list of constraints: `make_constriants(model, constraints)`. This should be separate from the function `lagrange_multipliers(model, constraints)` that is needed to implement #148 as this should return a list of suitably wrapped constraints that can be fed to `SLSQP`, whereas `lagrange_multipliers` should return a single model that can be fed to any optimizer.

pckroon:
Sounds good! Go for it! ;)

Issue #156, opened 2018-05-10
SLSQP should accept symfit-style constraints


tBuLi:
Currently, when using SLSQP directly one has to provide constraints in a sympy facion:

```python
def f(x, a, b):
    return a * x + b

def chi_squared(a, b):
    return np.sum((ydata - f(xdata, a, b))**2)

custom_minimize = SLSQP(
    chi_squared, 
    [a, b], 
    constraints=[{'type': 'eq', 'fun': lambda a, b: a + b - 1}]
)
fit_results = custom_minimize.execute()
```
which is obviously a lot harder then doing
```python
custom_minimize = SLSQP(chi_squared, [a, b], constraints=[Eq(a + b, 1)])
```
(And this is ignoring manually adding the jacobian of the constraint, which will ruin your day even more.)

This syntax should be supported, as constraints normally only concern `Parameter` objects, so why make it complicated?

pckroon:
Looks like a good idea. Is this a problem for SLSQP only?
And yes, *usually* constraints are only between `Parameter` objects, but maybe not always ("my function value needs to be larger than 0"). 
Also related, what does sympy do when you do `a + b == 1`?

tBuLi:
Currently it is SLSQP only, but I agree that it should be added in such a way that it also works for other `ScipyMinimize` children that support constraints.

I just noticed actually that `ScipyMinimize` already has a classmethod `ScipyMinimize.scipy_constraints` for precisely this, which we apparently forgot to plug in for SLSQP while making 0.4.0. So this shouldn't be hard to fix.

I agree with your comment about constraints not always involving just Parameters, but in those cases you can still do the manual labor yourself. And those kinds of constraints do not work in a symbolic context yet, since you would then need to sum over the data points to have it return a scalar and such operations are not currently supported. However, they will be in 1.0.0.

`a + b == 1` in sympy performs a symbolical equality check, and would therefore return `False`. It does not return a `Eq(a + b, 1)` instance. Unfortunate, but not worth hacking. Believe me, I tried ;). You are btw allowed to say `a + b >= 1`, that does return a `Ge(a + b, 1)` object.

pckroon:
Sounds more like ScipyMinimize should call that on all it's constraints on __init__. That way it's dealt with across the board.

Too bad about the sympy equality thing. It makes sense though.

PR #195, opened 2018-10-31
Remove wrapping from Objectives

tBuLi:
This PR makes a couple of big changes to the objective/minimizer dynamic which I think will make it easier to add certain features in the future.

- Constraints are now no longer partialed with all the data before being fed to the minimizers by `Fit`, but instead turned into a `MinimizeModel` objective, since that's conceptually what they do. This the flow of information a lot clearer, especially for ScipyMinimize subclasses.
This also takes a significant step towards fixing #154, in principle clearing the way for normal Models to be used as constraints.
- `BaseObjective` subclasses now have an `invariant_kwargs` cached property, which returns all the data and fixed parameters demanded by its `.model`'s signature. This prepares the way to #194, creating a place where invariant quantities can be found and stored. Also, this means fixed parameters no longer have to partialed away, as they can be seen as invariants.
- More controversial, I think that the `list2kwargs` decorator should be removed, and instead `BaseObjective` subclasses should just have an extra argument, `ordered_parameters`, from which the parameters provided can be turned into kwargs internally. The reason for this is that it makes the minimizer code a lot cleaner and easier to debug due to way less partialing and wrapping, and it doesn't take anything away from the API. Scipy, still our main supplier of minimizers, always wants to use such an argument, and if in the future we have a minimizer which doesn't then the preferred `**parameters` are still there.
- This removal of `list2kwargs` does create one new problem: when using a custum objective and feeding it to e.g. the SLSQP class, the signature of such a model has to be the scipy one, not the prefered keyword argument one. This can be solved I believe, by realizing that the new `CallableNumericalModel` was created exactly for this purpose: to give a sensible interface between a custom callable and the `symfit` API. Therefore, I suggest that `BaseMinimizer` should check if it was provided with something which is not an instance of `BaseObjective`, and if not we transform that into `MinimizeModel(CallableNumericalModel(custom_objective))` instead.

### DIFF
--- a/symfit/core/fit.py
+++ b/symfit/core/fit.py
@@ -240,7 +240,6 @@ class BaseNumericalModel(BaseModel):
     """
     def __init__(self, model, independent_vars, params):
         """
-
         :param model: dict of ``callable``, where dependent variables are the
             keys. If instead of a dict a (sequence of) ``callable`` is provided,
             it will be turned into a dict automatically.
@@ -294,7 +293,8 @@ class BaseCallableModel(BaseModel):
     """
     def eval_components(self, *args, **kwargs):
         """
-        :return: lambda functions of each of the components in model_dict, to be used in numerical calculation.
+        :return: lambda functions of each of the components in model_dict, to be
+            used in numerical calculation.
         """
         return [expr(*args, **kwargs) for expr in self.numerical_components]
 
@@ -455,7 +455,7 @@ class CallableModel(BaseCallableModel):
     def numerical_components(self):
         """
         :return: lambda functions of each of the analytical components in
-        model_dict, to be used in numerical calculation.
+            model_dict, to be used in numerical calculation.
         """
         return [sympy_to_py(expr, self.independent_vars, self.params) for expr in self.values()]
 
@@ -479,9 +479,10 @@ class Model(CallableModel):
     @property
     def jacobian(self):
         """
-        :return: Jacobian 'Matrix' filled with the symbolic expressions for all the partial derivatives.
-          Partial derivatives are of the components of the function with respect to the Parameter's,
-          not the independent Variable's.
+        :return: Jacobian 'Matrix' filled with the symbolic expressions for all
+            the partial derivatives. Partial derivatives are of the components
+            of the function with respect to the Parameter's, not the independent
+            Variable's.
         """
         return [[sympy.diff(expr, param) for param in self.params] for expr in self.values()]
 
@@ -495,7 +496,8 @@ class Model(CallableModel):
     @property
     def chi(self):
         """
-        :return: Symbolic Square root of :math:`\\chi^2`. Required for MINPACK optimization only. Denoted as :math:`\\sqrt(\\chi^2)`
+        :return: Symbolic Square root of :math:`\\chi^2`. Required for MINPACK
+            optimization only. Denoted as :math:`\\sqrt(\\chi^2)`
         """
         return sympy.sqrt(self.chi_squared)
 
@@ -503,7 +505,8 @@ class Model(CallableModel):
     def chi_jacobian(self):
         """
         Return a symbolic jacobian of the :math:`\\sqrt(\\chi^2)` function.
-        Vector of derivatives w.r.t. each parameter. Not a Matrix but a vector! This is because that's what leastsq needs.
+        Vector of derivatives w.r.t. each parameter. Not a Matrix but a vector!
+        This is because that's what leastsq needs.
         """
         jac = []
         for param in self.params:
@@ -535,28 +538,32 @@ class Model(CallableModel):
     @cached_property
     def numerical_jacobian(self):
         """
-        :return: lambda functions of the jacobian matrix of the function, which can be used in numerical optimization.
+        :return: lambda functions of the jacobian matrix of the function, which
+            can be used in numerical optimization.
         """
         return [[sympy_to_py(partial_dv, self.independent_vars, self.params) for partial_dv in row] for row in self.jacobian]
 
     @cached_property
     def numerical_chi_squared(self):
         """
-        :return: lambda function of the ``.chi_squared`` method, to be used in numerical optimisation.
+        :return: lambda function of the ``.chi_squared`` method, to be used in
+            numerical optimisation.
         """
         return sympy_to_py(self.chi_squared, self.vars, self.params)
 
     @cached_property
     def numerical_chi(self):
         """
-        :return: lambda function of the ``.chi`` method, to be used in MINPACK optimisation.
+        :return: lambda function of the ``.chi`` method, to be used in MINPACK
+            optimisation.
         """
         return sympy_to_py(self.chi, self.vars, self.params)
 
     @cached_property
     def numerical_chi_jacobian(self):
         """
-        :return: lambda functions of the jacobian of the ``.chi`` method, which can be used in numerical optimization.
+        :return: lambda functions of the jacobian of the ``.chi`` method, which
+            can be used in numerical optimization.
         """
         return [sympy_to_py(component, self.vars, self.params) for component in self.chi_jacobian]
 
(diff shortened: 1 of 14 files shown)


pckroon:
This reads like a sane idea. I still have to look at the code though. That said, for #189 I also went through the inheritance structure of minimizers. Could you implement the improvements I made there here?

pckroon:
Another thing we can seriously consider is removing the ScipyConstrainedMinimize baseclass. There are only 3 minimizers in scipy that can deal with constraints (COBYLA, SLSQP and trust-constr), and all three want something different. COBYLA doesn't like jacobians, SLSQP needs them, and trust-constr needs a different format altogether. So just give each of these wrappers some code that wraps the symfit constraints.

symfit/core/fit.py
```diff
@@ -193,6 +193,13 @@ def shared_parameters(self):
             else:
                 return False
 
+    @property
+    def unfixed_params(self):
```
pckroon: `free_params`?

symfit/core/minimizers.py
```diff
@@ -31,7 +32,18 @@ def __init__(self, objective, parameters):
         """
         self.parameters = parameters
         self._fixed_params = [p for p in parameters if p.fixed]
-        self.objective = partial(objective, **{p.name: p.value for p in self._fixed_params})
+        if isinstance(objective, BaseObjective):
+            self.objective = objective
+        else:
+            # Minimize the provided custom objective instead. This why want to
+            # minimize a CallableNumericalModel, thats what they are for.
+            from .fit import CallableNumericalModel
```
pckroon: Import higher up

symfit/core/objectives.py
```diff
@@ -57,50 +57,84 @@ def sigma_data(self):
...
+            self.model.__signature__.parameters if p in data_by_name}
+        )
+        return kwargs
+
+    def _clear_cache(self):
+        """
+        Clear cached properties. Should preferably be called after every
+        minimization precedure to clean the pallette.
+        """
+        del self.invariant_kwargs
```
pckroon: This should be a general method on one of the most basic classes that wipes *all* cached properties. Also, did you test this? It looks funny to me.

symfit/core/minimizers.py
tBuLi: Unfortunately it has to be here to prevent cyclic imports. Perhaps we can consider moving models to a seperate module in a seperate PR and then this import can be performed normally.

symfit/core/fit.py
tBuLi: That might be better, thanks

symfit/core/objectives.py
tBuLi: I'm still looking for a good place to call this after any execute, so suggestions are welcome.

The `del` indeed looks weird but it is correct, when you call del on a `cached_property` it is overloaded to delete the associated cache, not the whole property.

symfit/core/minimizers.py
pckroon: Makes sense. I am in favour of splitting some files, but that is indeed for later

symfit/core/minimizers.py
```diff
@@ -111,6 +125,7 @@ def __init__(self, *args, **kwargs):
         self._pickle_kwargs['jacobian'] = self.jacobian
 
         if self.jacobian is not None:
+            self.jacobian = self._baseobjective_from_callable(self.jacobian)
```
pckroon: The Jacobian is not an objective is it? If it needs wrapping, Objective should do that I think.

symfit/core/minimizers.py
```diff
@@ -361,18 +376,43 @@ def scipy_constraints(self, constraints):
...
         """
         cons = []
         types = {  # scipy only distinguishes two types of constraint.
             sympy.Eq: 'eq', sympy.Ge: 'ineq',
         }
 
-        for minimize_constraint in constraints:
-            constraint_type = minimize_constraint.model.constraint_type
+        for constraint in constraints:
+            if isinstance(constraint, MinimizeModel):
```
pckroon: I'd argue that if you want to call a Minimizer directly, you should do all the packaging and wrapping of input yourself. Especially since we disrecommend doing that in the first place. We can still consider providing a helper function that does it for you (which Fit also uses), but that would be pure benevolence on our part.

tests/test_minimize.py
```diff
@@ -169,8 +169,9 @@ def test_basinhopping(self):
         res = basinhopping(func, x0, minimizer_kwargs={"method": "BFGS"}, niter=200)
         np.random.seed(555)
         x, = parameters('x')
-        fit = BasinHopping(func, [x])
-        fit_result = fit.execute(minimizer_kwargs={"method": "BFGS", 'jac': False}, niter=200)
+        fit = BasinHopping(func, [x], local_minimizer=BFGS)
```
pckroon:
```
fit = Fit(func, minimizer=BasinHopping)
fit.execute(local_minimizer=BFGS)
```

tests/test_minimize.py
```diff
@@ -200,12 +201,21 @@ def jac2d_symfit(x1, x2):
...
+        with self.assertRaises(TypeError):
+            fit = BasinHopping(
+                func2d_symfit, [x1, x2],
+                local_minimizer=NelderMead(func2d_symfit, [x1, x2],
+                                           jacobian=jac2d_symfit)
+            )
+        fit = BasinHopping(
+            func2d_symfit, [x1, x2],
+            local_minimizer=BFGS(func2d_symfit, [x1, x2], jacobian=jac2d_symfit)
+        )
```
pckroon: Same as above. Use Fit.

symfit/core/minimizers.py
tBuLi: This is done in case a custom callable is given as a jacobian, just like for the objective. This happens for example when doing things like:
```python
def func2d_symfit(x1, x2):
    f = np.cos(14.5 * x1 - 0.3) + (x2 + 0.2) * x2 + (x1 + 0.2) * x1
    return f

def jac2d_symfit(x1, x2):
    df = np.zeros(2)
    df[0] = -14.5 * np.sin(14.5 * x1 - 0.3) + 2. * x1 + 0.2
    df[1] = 2. * x2 + 0.2
    return df

x1, x2 = parameters('x1, x2')
fit = BFGS(func2d_symfit, [x1, x2], jacobian=jac2d_symfit)
```

symfit/core/minimizers.py
tBuLi: So that would mean that we just assume a `MinimizeModel` instance here for constraints, as `Fit` provides, and take it from there?

I do tend to agree, but I also do think that the code example in #156 should work, it just looks so natural, don't you think? But even if we keep that one, there's one of the conditions here we could skip. I also definitely agree that people should be encouraged as much as possible to just use Fit.

tests/test_minimize.py
tBuLi: This is meant as tests on the Minimizer directly, not if it plays nice with Fit. That's a couple of lines down ;).

symfit/core/minimizers.py
pckroon: Ok, good point.

symfit/core/minimizers.py
pckroon: The more I think about it, the stronger I am of the opinion that if people choose not to use `Fit`, they should wrap their own things. That way 1) our code stays cleaner, since we *know* what kind of input we get, and 2) people still have access to the powertools. 
The way you propose here provides no additional *functionality* over people calling Fit, or wrapping their constraints in a sane way.

symfit/core/minimizers.py
pckroon: On second thought I still don't like it. People calling minimizers should do their own wrapping. This means either 1) creating a `CallableNumericalModel` which takes a `jacobian` argument, or 2) calling `Fit`, which should then also take a `jacobian` argument.

symfit/core/minimizers.py
```diff
@@ -31,11 +32,36 @@ def __init__(self, objective, parameters):
         """
         self.parameters = parameters
         self._fixed_params = [p for p in parameters if p.fixed]
-        self.objective = partial(objective, **{p.name: p.value for p in self._fixed_params})
+        self.objective = self._baseobjective_from_callable(objective)
```
pckroon: My more extensive comments elsewhere also means that this should be moved to `Fit`.

symfit/core/minimizers.py
pckroon: But of course, we could provide some convenience functions to help them

PR #196, opened 2018-11-01
Hessian

pckroon:
By request.
Closes #190.

PR #199, opened 2018-11-03
Hessian and many other steps towards 0.5.0

tBuLi:
Starting from #195 and #196, this completes the job of adding hessians to our models and objectives in the new structure initiated with #195.

However, it became much more than just that because of the realization that Hessians quite easily destroy the shape of our models, because deriving twice wrt a parameter seriously increases the likelihood of getting 0. When this happens, we were no longer able to infer the output shape in the previous scheme.

This PR therefore radically changes the approach to shaping:
- Models can not know about the shape of the dependent data, only the objectives can. So for example, when averaging `{y: a}` with `y` a dataset and `a` just a constant, the model is unable to know the shape of the result, even-though we need an array with the same shape as `y`. This is now done in the objective in `_shape_dependent_data`, which is called on all model results before continuing.
- When the model does have vars, e.g. `{y: a * x}`, then the jacobian and hessian lose the shape information: `{D(y, a): a}` and `{D(y, a, a): 0}`. This is bad, because it destroys broadcasting when going to numpy arrays for models less trivial then this one. This PR proposes the following solution: replace `x` by `x * OnesLike(x)`, where `OnesLike` is a new sympolic object which gets printed into `np.ones_like`.
This solves the problem because of the properties of `OnesLike` and its sibling `ZerosLike` under differentiation, addition etc. In the simple example above:
```python
{y: a * x * OnesLike(x, a)}
{D(y, a): a * OnesLike(x, a)}
{D(y, a, a): ZerosLike(x, a)}
```
(all parameters have to arguments of this object because otherwise sympy recognizes that this derivative is trivially zero)

Other changes:
- Hessian is used for covariance matrix in LogLikelihood
- 

Things to be considered:
- Hessian should always be used as the covariance matrix if possible.
- Better simplification rules for OnesLike, ZerosLike and VarOnesLikeVar
- Discussion about the shapes of Jacobians and Hessians needs to be had. I think that for objectives, __call__ should return a scalar, eval_jacobian a vector, and eval_hessian a matrix. This is also the way I implemented it now. However, I feel that if you feed the symbolic equivalent to MinimizeModel, we should get the same shapes.
```python
chi2_exact = Model({X2: Sum(((a * x ** 2 + b * x) - y) ** 2, i)})
eval_exact = chi2_exact(x=xdata, y=ydata, a=2, b=3)                         # shape: (1, 1)
jac_exact = chi2_exact.eval_jacobian(x=xdata, y=ydata, a=2, b=3)  # shape: (1, 2, 1)
hess_exact = chi2_exact.eval_hessian(x=xdata, y=ydata, a=2, b=3)  # shape: (1, 2, 2, 1)
```
This first 1 is justified, it is due to the fact that models are always iterable over the components, even scalar models. But the last one should not be there I feel. This only appears because of our convention that the elements should always be at least 1d arrays, even if it is a scalar. If we get rid of this convention, the output from `MinimizeModel(chi2_exact, data={})` would be identical to that of `LeastSquares` of the same function, which is something which makes a lot of sense to me. However, I don't think `scipy` cares either way so I left it like this for now.

### DIFF
--- a/symfit/core/argument.py
+++ b/symfit/core/argument.py
@@ -115,3 +115,9 @@ class Variable(Argument):
     # Variable index to be assigned to generated nameless variables
     _argument_name = 'var'
     __slots__ = ()
+
+    def _numpycode(self, printer):
+        return printer.doprint(self.name)
+
+    _lambdacode = _numpycode
+    _sympystr = _numpycode
--- a/symfit/core/fit.py
+++ b/symfit/core/fit.py
@@ -490,6 +490,15 @@ class Model(CallableModel):
         return [[sympy.diff(expr, param) for param in self.params] for expr in self.values()]
 
     @property
+    def hessian(self):
+        """
+        :return: Hessian 'Matrix' filled with the symbolic expressions for all the partial derivatives.
+          Partial derivatives are of the components of the function with respect to the Parameter's,
+          not the independent Variable's.
+        """
+        return [[[sympy.diff(diff1, param2) for param2 in self.params] for diff1 in comp] for comp in self.jacobian]
+
+    @property
     def chi_squared(self):
         """
         :return: Symbolic :math:`\\chi^2`
@@ -547,6 +556,20 @@ class Model(CallableModel):
         return [[sympy_to_py(partial_dv, self.independent_vars, self.params) for partial_dv in row] for row in self.jacobian]
 
     @cached_property
+    def numerical_hessian(self):
+        """
+        :return: lambda functions of the Hessian matrix of the function, which can be used in numerical optimization.
+        """
+        return [
+                [
+                 [
+                  sympy_to_py(diff2, self.independent_vars, self.params)
+                  for diff2 in diff1
+                 ] for diff1 in comp
+                ] for comp in self.hessian
+               ]
+
+    @cached_property
     def numerical_chi_squared(self):
         """
         :return: lambda function of the ``.chi_squared`` method, to be used in
@@ -586,41 +609,36 @@ class Model(CallableModel):
             [partial_dv(*args, **kwargs) for partial_dv in row ] for row in self.numerical_jacobian
         ]
         for idx, comp in enumerate(jac):
-            # Find out how many datapoints this component has. We need to do
-            # this with a try/except, since partial_derivative can be a number or
-            # a sequence. We ultimately want to make sure every evey component
-            # has this size, so component of the jacobian can be contracted properly.
-            data_len = 1
-            for partial_derivative in comp:
-                if hasattr(partial_derivative, 'shape') and partial_derivative.shape:
-                    # Last line is to descriminate against numpy.float of shape (,)
-                    shape = partial_derivative.shape
-                else:
-                    try:
-                        shape = len(partial_derivative)
-                    except TypeError: # Not iterable, so assume number
-                        shape = 1
-                if isinstance(shape, tuple):
-                    if isinstance(data_len, tuple):
-                        if len(shape) > len(data_len):
-                            data_len = shape
-
-                    else:
-                        data_len = shape
-                elif isinstance(data_len, tuple):
-                    # data_len is a tuple, but shape isn't. Prefer the tuple.
-                    pass
-                else:
-                    data_len = max(shape, data_len)
-            # And make everything in this component the same size, since some are
-            # numbers.
-            for jdx, partial_derivative in enumerate(comp):
-                # This is a no-op for elements of size `longest`.
-                jac[idx][jdx] = np.ones(data_len) * partial_derivative
-            # And lastly, turn jac into a list of 2D numpy arrays of shape
-            # (Nparams, Ndata)
-            jac[idx] = np.array(jac[idx], dtype=float)
+            shapes = [np.atleast_1d(diff).shape for diff in comp]
+            # Funny key so that higher dimensional data > lower dimensional
+            # data
+            largest = max(shapes, key=lambda s: (len(s), *s))
+            ones = np.ones(largest, dtype=float)
+            data = np.array([ones * diff for diff in comp])
+            jac[idx] = data.reshape((len(comp), *largest))
         return jac
+
+    def eval_hessian(self, *args, **kwargs):
+        """
+        :return: Hessian evaluated at the specified point.
+        """
+        hess = [
+                [
+                 [
+                  diff2(*args, **kwargs) for diff2 in diff1
+                 ] for diff1 in comp
+                ] for comp in self.numerical_hessian
+               ]
+        for idx, comp in enumerate(hess):
+            shapes = [np.atleast_1d(d2).shape for d1 in comp for d2 in d1]
+            # Funny key so that higher dimensional data > lower dimensional
+            # data
+            largest = max(shapes,
+                          key=lambda s: (len(s), *s))
+            ones = np.ones(largest, dtype=float)
+            data = np.array([ones * d2 for d1 in comp for d2 in d1])
+            hess[idx] = data.reshape((len(comp), len(comp), *largest))
+        return hess
 
 
 class TaylorModel(Model):
--- a/symfit/core/fit.py
+++ b/symfit/core/fit.py
@@ -299,7 +299,7 @@ class BaseCallableModel(BaseModel):
         :return: lambda functions of each of the components in model_dict, to be
             used in numerical calculation.
         """
-        return [expr(*args, **kwargs) for expr in self.numerical_components]
+        return [np.atleast_1d(expr(*args, **kwargs)) for expr in self.numerical_components]
 
     @abstractmethod
     def numerical_components(self):
--- a/symfit/core/fit.py
+++ b/symfit/core/fit.py
@@ -1504,12 +1504,14 @@ class Fit(HasCovarianceMatrix):
                 minimizer_options['jacobian'] = self.objective.eval_jacobian
 
         if issubclass(minimizer, ConstrainedMinimizer):
-            # Minimizers are agnostic about data, they just know about
-            # objective functions. So we partial away the data at this point.
-            minimizer_options['constraints'] = [
-                MinimizeModel(constraint, data=self.data)
-                for constraint in self.constraints
-            ]
+            # set the constraints as MinimizeModel. The dependent vars of the
+            # constraint are set to None since their value is irrelevant.
+            constraint_objectives = []
+            for constraint in self.constraints:
+                data = self.data  # No copy, share state
+                data.update({var: None for var in constraint.dependent_vars})
+                constraint_objectives.append(MinimizeModel(constraint, data))
+            minimizer_options['constraints'] = constraint_objectives
         return minimizer(self.objective, self.model.params, **minimizer_options)
 
     def _init_constraints(self, constraints):
--- a/symfit/core/fit.py
+++ b/symfit/core/fit.py
@@ -994,9 +994,8 @@ class HasCovarianceMatrix(TakesData):
         if isinstance(self.objective, LogLikelihood):
             # Loglikelihood is a special case that needs to be considered
             # separately, see #138
-            jac = self.objective.eval_jacobian(apply_func=lambda x: x, **key2str(best_fit_params))
-            cov_matrix_inv = np.tensordot(jac, jac, (range(1, jac.ndim), range(1, jac.ndim)))
-            cov_mat = np.linalg.inv(cov_matrix_inv)
+            hess = self.objective.eval_hessian(**key2str(best_fit_params))
+            cov_mat = np.linalg.inv(hess)
             return cov_mat
         try:
             if len(set(arr.shape for arr in self.sigma_data.values())) == 1:
--- a/symfit/core/fit.py
+++ b/symfit/core/fit.py
@@ -986,13 +986,13 @@ class HasCovarianceMatrix(TakesData):
             return np.array(
                 [[float('nan') for p in self.model.params] for p in self.model.params]
             )
-        if isinstance(self.objective, LogLikelihood):
-            # Loglikelihood is a special case that needs to be considered
-            # separately, see #138
-            hess = self.objective.eval_hessian(**key2str(best_fit_params))
-            cov_mat = np.linalg.inv(hess)
-            return cov_mat
         try:
+            if isinstance(self.objective, LogLikelihood):
+                # Loglikelihood is a special case that needs to be considered
+                # separately, see #138
+                hess = self.objective.eval_hessian(**key2str(best_fit_params))
+                cov_mat = np.linalg.inv(hess)
+                return cov_mat
             if len(set(arr.shape for arr in self.sigma_data.values())) == 1:
                 # Shapes of all sigma data identical
                 return self._cov_mat_equal_lenghts(best_fit_params=best_fit_params)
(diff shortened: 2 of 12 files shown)


tBuLi:
To answer one of my own questions: scipy indeed doesn't care about the trailing dimension, so both `()` and `(1,)` are scalar, `(m,)` and `(m, 1)` are vectors, etc. This means that the analytical model plus MinimizeModel immediately works and gives the same result, so I'm happy. (I'll add this as a test) So as far as I'm concerned we can keep the current convention.

tBuLi:
@pckroon I agree with your description for the model, but for the objective you made a mistake. A call to an objective should return a scalar, so that's `(1,)`, not `(n, 1)`. Same goes for the jacobian and hessian, there should be no `n`. Otherwise I agree with you.

After a good nights rest I realize that perhaps I overdid it by introducing the OnesLike objects, and we can keep your previous reshaping code. I now think the reason the tests are running for hessians is not the OnesLike, but my changes on reshaping on the objective side, where we make sure the shape matches against that of the dependent data.

I'll update and report back on that later.

pckroon:
That's what you get when I post comments last thing  before I go to bed ;)
And yes, GH travelled through time.

If you leave the model hessian broadcasting code as I put it, you end up with hessians of shape `(k, k, 1 or m)` per component, with m being the shape of the independent data involved in that component. I'll wait for travis to finish, then I'll go through the changes.

symfit/core/fit.py
```diff
@@ -1491,12 +1545,14 @@ def _init_minimizer(self, minimizer, **minimizer_options):
...
-            minimizer_options['constraints'] = [
-                MinimizeModel(constraint, data=self.data)
-                for constraint in self.constraints
-            ]
+            # set the constraints as MinimizeModel. The dependent vars of the
+            # constraint are set to None since their value is irrelevant.
+            constraint_objectives = []
+            for constraint in self.constraints:
+                data = self.data  # No copy, share state
+                data.update({var: None for var in constraint.dependent_vars})
```
pckroon: Are you very very sure about the no copy?

symfit/core/objectives.py
```diff
@@ -69,7 +69,34 @@ def __call__(self, ordered_parameters=[], **parameters):
...
+
+        :param model_output: Output of a call to model
+        :param param_level: indicates how many parameter dimensions should be
+            added to the shape. 0 for __call__, 1 for jac, 2 for hess.
+        :return: ``model_output`` reshaped to ``dependent_data``'s shape, if
+            possible.
+        """
+        shaped_result = []
+        n_params = len(self.model.params)
+        for dep_data, component in zip(self.dependent_data.values(), model_output):
```
pckroon: Are these guaranteed to have the same order on all versions of Python?

symfit/core/objectives.py
```diff
@@ -69,7 +69,34 @@ def __call__(self, ordered_parameters=[], **parameters):
...
+            if dep_data is not None:
+                if dep_data.shape == component.shape:
+                    shaped_result.append(component)
+                else:
+                    # Let numpy deal with all the broadcasting
+                    shape = param_level * [n_params] + list(dep_data.shape)
+                    shaped_result.append(np.broadcast_to(component, shape))
```
pckroon: broadcast_to is a nice find! I'll be sure to remember it.

symfit/core/objectives.py
```diff
@@ -113,7 +140,49 @@ def eval_jacobian(self, ordered_parameters=[], **parameters):
...
+        :return: evaluated hessian
+        """
+        parameters.update(dict(zip(self.model.free_params, ordered_parameters)))
+        parameters.update(self.invariant_kwargs)
+        result = self.model.eval_hessian(**key2str(parameters))
+        return self._shape_dependent_data(result, param_level=2)
+
+
+@add_metaclass(abc.ABCMeta)
+class HessianObjective(GradientObjective):
```
pckroon: There's 2 `HessianObjectives`.

symfit/core/objectives.py
```diff
@@ -213,21 +282,55 @@ def eval_jacobian(self, ordered_parameters=[], **parameters):
...
+            ordered_parameters, **parameters
+        )
+        evaluated_hess = super(LeastSquares, self).eval_hessian(
+            ordered_parameters, **parameters
+        )
 
-class LogLikelihood(GradientObjective):
+        result = 0
+        for var, f, jac_comp, hess_comp in zip(self.model, evaluated_func,
+                                               evaluated_jac, evaluated_hess):
```
pckroon: Same order?

symfit/core/objectives.py
```diff
@@ -213,21 +282,55 @@ def eval_jacobian(self, ordered_parameters=[], **parameters):
...
+            y = self.dependent_data[var]
+            sigma_var = self.model.sigmas[var]
+            if y is not None:
+                sigma = self.sigma_data[sigma_var]
+                p1 = hess_comp * ((y - f) / sigma**2)[np.newaxis, np.newaxis, ...]
+                # Outer product
+                p2 = np.einsum('i...,j...->ij...', jac_comp, jac_comp)
+                p2 / sigma[np.newaxis, np.newaxis, ...]**2
```
pckroon: `/=`? This is currently a noop

symfit/core/objectives.py
```diff
@@ -296,5 +429,14 @@ def eval_jacobian(self, ordered_parameters=[], **parameters):
...
             return np.array(evaluated_jac[0])
+        else:
+            return None
+
+    def eval_hessian(self, ordered_parameters=[], **parameters):
+        if hasattr(self.model, 'eval_hessian'):
+            evaluated_hess = super(MinimizeModel, self).eval_hessian(
+                ordered_parameters, **parameters
+            )
+            return np.array(evaluated_hess[0])
```
pckroon: Do we check anywhere that is is in fact a single component model?

symfit/core/printing.py
```diff
@@ -0,0 +1,103 @@
...
+        result = (num_syms, self._not_supported, "\n".join(lines))
+    self._not_supported = set()
+    self._number_symbols = set()
+    return result
+
+CodePrinter.doprint = doprint
+
+#########################################################
+# End of monkeypatch                                    #
+#########################################################
```
pckroon: I assume most of this came from sympy?

tests/test_objectives.py
```diff
@@ -7,12 +7,24 @@
...
 from symfit.core.objectives import (
     VectorLeastSquares, LeastSquares, LogLikelihood, MinimizeModel
 )
 from symfit.core.fit_results import FitResults
+from symfit.core.printing import SymfitNumPyPrinter
+from symfit.distributions import Exp
+
+# Overwrite the way Sum is printed by numpy just while testing. Is not
+# general enough to be moved to SymfitNumPyPrinter, but has to be used
+# in this test.
```
pckroon: Why does it have to be used?

symfit/core/argument.py
```diff
@@ -115,3 +116,183 @@ class Variable(Argument):
...
+                if set(self.args) == set(constant_like.args):
+                    return other
+
+        return super(ZerosLike, self).__add__(other)
+
+    def fdiff(self, argindex=1):
+        """
+        Returns the first derivative of the function.
+        """
+        return ZerosLike(*self.args)
```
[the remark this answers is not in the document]
tBuLi: good point

symfit/core/argument.py
```diff
@@ -115,3 +116,183 @@ class Variable(Argument):
...
+            return ZerosLike(*self.args)
+
+
+def _simplify_constantlike(expr, reduce_shape=False):
+    """
+    Simple arithmetic involving ConstantLike objects can be greatly simplified,
+    but that is not done by default because e.g. sympy.diff does not explicitly
+    call Symbol.__mul__ but instead creates a Mul object. This function loops
+    through the expressions tree and makes sure basic math operations are
+    performed explicitly so the simplification kicks in.
```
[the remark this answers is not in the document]
tBuLi: Unfortunatelly, no. sympy's normal simplify functionality will not explicitelly perform operation like `*`, so the simplifying statements in `__mul__` never get called. From reading a bit this seems for good reasons, but here it's definitely preferred.

symfit/core/fit.py
```diff
@@ -308,6 +312,26 @@ def numerical_components(self):
...
 
+    @property
+    def _shaped_values(self):
+        """
+        Returns self.values(), i.e. the components, but with vars replaced by
+        OnesLike(var) * var to safegaurd the shape.
+        """
+        shaped_expr = self.model_dict.copy()
+
+        for y in shaped_expr:
```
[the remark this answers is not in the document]
tBuLi: Doesn't work because I want to modify the dict values.

symfit/core/fit.py
```diff
@@ -479,15 +504,49 @@ class Model(CallableModel):
...
+            respect to the Parameter's, not the independent Variable's.
         """
-        return [[sympy.diff(expr, param) for param in self.params] for expr in self.values()]
+        return [[[_simplify_constantlike(sympy.diff(partial_dv, param))
+                    for param in self.params]
+                for partial_dv in comp]
+            for comp in self._jacobian(self._shaped_values)]
+
+    @property
+    def hessian(self):
```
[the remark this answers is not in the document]
tBuLi: I should've just done git push -f and nuked everything, being forced to pull first created a mess so it seems.

symfit/core/objectives.py
tBuLi: Yes, I use always use protection: everything is an OrderedDict.

symfit/core/objectives.py
tBuLi: good find

symfit/core/objectives.py
tBuLi: From the docstring:
```python
    """
    Objective to use when the model itself is the quantity that should be
    minimized. This is only supported for scalar models.
    """
```
And that rule is enforced in __init__ ;)

symfit/core/printing.py
tBuLi: Just a copy paste indeed. But I now realize that this can be done much better by changing to a decorator for the two items that caused problems. The problem was that they called `del` on two things that should not be deleted. But if I make a property I can ignore the delete in the first place.

tBuLi: And then it is more robust if there are other changes to this code.

tests/test_objectives.py
tBuLi: In order to build the analytical chi2 function we need to sum over `i`, but that is not printed currently by the numpy printer. And the solution here of replacing by `np.sum` is in general definitely not correct since that sums over all dimensions, not just that specified in `Sum`.

tBuLi:
After playing around some more, your previous changes to the shapes returned by models combined with my new changes to objectives give the right results, all tests run locally apart from one of my new ones.

But it is an import one to discuss:
```python
model = Model({y: a * x**2 + b * x})
self.assertEqual(model(x=xdata, a=2, b=3)[0].shape, ydata.shape)
self.assertEqual(model.eval_jacobian(x=xdata, a=2, b=3)[0].shape, (2, 100))
self.assertEqual(model.eval_hessian(x=xdata, a=2, b=3)[0].shape, (2, 2, 100))
``` 
The last line fails, giving `(2, 2, 1)` instead. Now, this does not stop the fits from working because the objective will correct for this based on the dependent data. However, philosophically, are we ok with that?

If we are, the `OnesLike` are unnecessary. and should be removed, or at least not automatically used by symfit. Perhaps they can be left so users can use them explicitly if they want.

tests/test_objectives.py
pckroon: You could do a sum by axis, but that'll be iffy to implement correctly I guess. Also, add that explanation to the comment

pckroon:
I am ok with that because 1) the broadcasting works, and 2) the hessian *is* independent of the data. Comment (don't remove) the offending assert and add a comment.
This may come back to bite us at some point, but we;ll fix it then.

symfit/core/fit.py
```diff
@@ -586,42 +608,34 @@ def eval_jacobian(self, *args, **kwargs):
...
-                jac[idx][jdx] = np.ones(data_len) * partial_derivative
-            # And lastly, turn jac into a list of 2D numpy arrays of shape
-            # (Nparams, Ndata)
-            jac[idx] = np.array(jac[idx], dtype=float)
+            shapes = [np.atleast_1d(diff).shape for diff in comp]
+            # Funny key so that higher dimensional data > lower dimensional
+            # data
+            largest = max(shapes, key=lambda s: [len(s)]+list(s))
+            ones = np.ones(largest, dtype=float)
+            data = np.array([ones * diff for diff in comp])
```
pckroon: These `ones*diff` calls should probably be changed to `np.broadcast_to`

symfit/core/fit.py
```diff
@@ -586,42 +608,34 @@ def eval_jacobian(self, *args, **kwargs):
...
+                for row in comp]
+            for comp in self.numerical_hessian]
+
+        for idx, comp in enumerate(hess):
+            shapes = [np.atleast_1d(d2).shape for d1 in comp for d2 in d1]
+            # Funny key so that higher dimensional data > lower dimensional
+            # data
+            largest = max(shapes, key=lambda s: [len(s)]+list(s))
+            ones = np.ones(largest, dtype=float)
+            data = np.array([ones * d2 for d1 in comp for d2 in d1])
```
pckroon: These `ones*diff` calls should probably be changed to `np.broadcast_to`

pckroon:
Code is almost good to go as far as I can see. There's one more missing comment in the tests (about the Sum printer), and I want to be very sure we don't need `self.data.copy` in the constraints

pckroon:
Once those are addressed and the tests are fixed it's good.

tBuLi:
I'll resolve the other comments. About the `self.data` copy or not to copy, thats related to our discussion in #194. My thinking is that after each call to a model by it's objective, we could make the objective store that output in `self.data`. That way, this output is also available to any constraints initiated by the same `Fit` in their objective.

For example:
```
i = Idx('i')
x = MatrixSymbol('x', i, 1)
y = MatrixSymbol('y', i, 1)
model = {y: a * x**2}
constraints = [Eq(Sum(y[i, 0], i), 1)]
```
The constraint could then retrieve the value of `y` at every evaluation from `self.data`.

We have to think if we cannot just add the call result to `self.data` though, as that might overwrite the dependent data belonging to `y` and that would be bad. Ideas are welcome.

pckroon:
Clever, but dangerous. Food for thought for a next PR.

tBuLi:
I don't think it will result in catostrophe btw, and here's why:

In the objective we take the dependent data from `self.dependent_data`, which is a new OrderedDict generated from `self.data`. If we call this cached_property explicitely in init, we are sure that it is filled with the real data. In the rest of the objective, we always access dependent data trought this attribute, never through `self.data`. Then, after calling the model, we can store its result in `self.data` by simply updating the dict.

Then, in the `MinimizeModel` of the constraint, this output is now available, and since what was a dependent var for the model is an independent var for the constraint, it will access the model's output directly through the new `invariant_kwargs`, which takes every dataset needed from 'self.data' based on the signature.

symfit/core/fit.py
```diff
@@ -604,36 +604,32 @@ def eval_jacobian(self, *args, **kwargs):
...
         for idx, comp in enumerate(hess):
-            shapes = [np.atleast_1d(d2).shape for d1 in comp for d2 in d1]
-            # Funny key so that higher dimensional data > lower dimensional
-            # data
-            largest = max(shapes, key=lambda s: [len(s)]+list(s))
-            ones = np.ones(largest, dtype=float)
-            data = np.array([ones * d2 for d1 in comp for d2 in d1])
-            hess[idx] = data.reshape([len(comp), len(comp)] + list(largest))
+            hess[idx] = np.stack(np.broadcast_arrays(*comp))
+            print(hess[idx].shape)
```
pckroon: Erroneous print

Issue #39, opened 2015-12-07
Make model calls consistent


tBuLi:
Currently there are a couple of equivalent functions in `Model` and expr that do roughly the same. I should check to make sure that there is no overlap in their tasks. How I think it should work:

`Expr.__call__` evaluates a sympy `Expr` with given args and kwargs.
`Model.__call__` should evaluate each of the `Expr` in `model_dict`.
`Model.numerical_components` is thereby redundant and should be removed. There should be only one obvious way, and thats to call `Model`.

pckroon:
I would like the jacobian methods to this list. There are a few of them, and they do more or less the same thing, but differently, and it's not immediately clear from their names. There should be a clear distinction between methods that use the symbolic information (and are fast/cheap) and those that use numerical approximations (and are slow/expensive).

tBuLi:
This has by now been solved: models should be evaluated by calling them, but the lambdafied functions are stored in `numerical_components`. Additionally, jacobians and hessians are now seperate models which are available under `model.jacobian_model` and `model.hessian_model`. This structure allows for a lot more efficient code reuse.

PR #201, opened 2018-11-18
Matrix compatibility and interdependent components

### LINKED ISSUE #84: __str__ for vector models improvement
Upon printing a vector valued model, the component are each printed on a separate line. However, this way it could also be multiple normal models being printed. This printed lines have to be updated to reflect the fact that it is a vector model.

tBuLi:
This PR makes two big new changes: support for `sympy.MatrixExpr` as variables and as an indirect consequence it also adds component interdependence.

Consider e.g.
```python
a, b = parameters('a, b')
model_dict = {
    y: a**3 * x + b**2,
    z: y**2 + a * b
}
```

This could be a great asset when `y` is expensive to compute and used multiple times. In such a model, `y` is no longer a dependent variable, but rather an interdependent one.

The big challenge here was how to calculate the Jacobian and Hessian for such a model, since also there we would like `D(z, a)` to be a function of `D(y, a)`:
```python
{D(y, a): 3 * a**2 * x,
 D(y, b): 2 * b,
 D(z, a): b + 2 * y * D(y, a),
 D(z, b): a + 2 * y * D(y, b),
}
```

This PR so far addresses:

- [x] Introduce separate functions for building a model to represent the jacobian, hessian and least-squares.
- [x] Models now have an attribute `self.jacobian_model` and `self.hessian_model`, which are models to represent those quantities. `self.eval_[...]` calls the respective model and properly arranges the output. (The models just have a set of component representing the jacobian and hessian, so they need to be ordered properly.)
- [x] In addition to the previous point, evaluating models is now based purely on calling the `eval_components`, `eval_jacobian`, and `eval_hessian`. Unless changed by a subclass, `eval_components` calls `numerical_components`. The other two simply call `self.jacobian_model` or `self.hessian_model`.
- [x] All models now have an important new property: `connectivity_mapping`. This indicates for every component on which variables and parameters it depends. This is used in code generation when lambdifying symbolic expressions, such that every component is no longer assumed to depend on all independent variables and parameters, but is just provided with those it needs.

   For `BaseNumericalModel` subclasses such a `connectivity_mapping` should be provided, although the old syntax is also still supported.
- [x] Based on `connectivity_mapping`, a topological sort can be performed to find the right evaluation order of the components. This is stored in `ordered_symbols`. This added a dependency to the very lightweight package `toposort`.
- [x] In order to properly perform the chain rule we need to turn all variables into functions before calculating derivatives. To this end, there is now 'vars_as_functions' to provide a simple mapping between the two, and `function_dict`, which is equivalent to `model_dict` except that all variables have been turned into `Function` objects and the sorting is based on `ordered_symbols`, not alphabetical.
- [x] Removed a lot of commented out fossilized code which stemmed from a different era.
- [x] Constraints have been massively simplified, getting ever closer to the goal of eliminating them completely.
- [ ] Fixed #84, printing vector models in a more recognizable way. We could still consider printing the `function_dict` instead though.
- [ ] Move models to a separate file.
- [x] Rescale chi2 by 1/2 so we can always invert the Hessian immediately to get the covariance matrix, no worries about factors of 2


symfit/core/fit.py
```diff
@@ -133,23 +170,98 @@ def _init_from_dict(self, model_dict):
...
+        functions = {}
+        # vars first, then params, and alphabetically within each group
+        key = lambda arg: [isinstance(arg, Parameter), str(arg)]
+        for symbol in self.ordered_symbols:
+            if symbol in self.connectivity_mapping:
+                connections = self.connectivity_mapping[symbol]
+                # Replace the connection by it's function if possible
+                connections = [functions.get(connection, connection)
+                               for connection in connections]
+                connections = sorted(connections, key=key)
```
pckroon: This is melting my brain. `connections` is a list containing either Functions, Variables, or Parameters, right? How do Functions get sorted compared to Variables in this story?

symfit/core/fit.py
```diff
@@ -133,23 +170,98 @@ def _init_from_dict(self, model_dict):
...
+                functions[symbol] = sympy.Function(symbol.name)(*(connections))
```
pckroon: Lose the brackets around `connections`

symfit/core/fit.py
```diff
@@ -133,23 +170,98 @@ def _init_from_dict(self, model_dict):
...
+            can be evaluated in the correct order.
+
+            Within each group of equal priority symbols, we sort by the order of
+            the derivative.
+        """
+        key_func = lambda s: [isinstance(s, sympy.Derivative),
+                           isinstance(s, sympy.Derivative) and s.derivative_count]
+        symbols = []
+        for d in toposort(self.connectivity_mapping):
+            symbols.extend(sorted(d, key=key_func))
```
pckroon: `d` is a terrible variable name

symfit/core/fit.py
```diff
@@ -209,14 +321,20 @@ def __str__(self):
         template = "{}({}; {}) = {}"
         parts = []
         for var, expr in self.items():
+            params_sorted = sorted((x for x in self.connectivity_mapping[var]
+                                    if isinstance(x, Parameter)),
+                                   key=lambda x: x.name)
+            vars_sorted = sorted((x for x in self.connectivity_mapping[var]
+                                  if x not in params_sorted),
+                                 key=lambda x: x.name)
```
pckroon: Why not just take `self.params` and `self.independent_vars + self.interdependent_vars`?

symfit/core/fit.py
```diff
@@ -238,32 +356,59 @@ class BaseNumericalModel(BaseModel):
...
+            # Make model into a mapping if needed.
+            if not isinstance(model, Mapping):
+                try:
+                    iter(model)
+                except TypeError:
+                    model = [model]  # make model iterable
+
+                with warnings.catch_warnings():
+                    warnings.simplefilter("ignore")
+                    model = {Variable(): expr for expr in model}
```
pckroon: Can use a comment on why the ignore warnings, and *should* pick a name for the Variable

symfit/core/fit.py
```diff
@@ -238,32 +356,59 @@ class BaseNumericalModel(BaseModel):
...
+            warnings.warn(DeprecationWarning(
+                '`independent_vars` and `params` have been deprecated.'
+                ' Use `connectivity_mapping` instead.'
+            ))
+            self.independent_vars = sorted(independent_vars, key=str)
+            self.params = sorted(params, key=str)
+            self.connectivity_mapping = {var: set(independent_vars + params)
+                                         for var in model}
```
pckroon: Everything depends on everything? Can't we do better?

symfit/core/fit.py
```diff
@@ -238,32 +356,59 @@ class BaseNumericalModel(BaseModel):
...
+        elif connectivity_mapping:
+            self.connectivity_mapping = connectivity_mapping
+            if not isinstance(model, Mapping):
+                raise TypeError('Please provide the model as a mapping, '
+                                'corresponding to `connectivity_mapping`.')
+        else:
+            raise TypeError('Provide either `connectivity_mapping` (preferred) '
+                            'or `independent_vars` and `params` (deprecated).')
```
pckroon: Just remove the last bit

symfit/core/fit.py
```diff
@@ -238,32 +356,59 @@ class BaseNumericalModel(BaseModel):
...
-        set in __init__.
-
-        :param model_dict: dict of (dependent_var, callable) pairs.
-        """
-        sort_func = lambda symbol: str(symbol)
-        self.model_dict = OrderedDict(sorted(model_dict.items(), key=lambda i: sort_func(i[0])))
-        self.dependent_vars = sorted(model_dict.keys(), key=sort_func)
+    @property
+    def connectivity_mapping(self):
+        return self._connectivity_mapping
```
pckroon: Why is this a property?

symfit/core/fit.py
```diff
@@ -335,10 +497,18 @@ def __call__(self, *args, **kwargs):
...
+        return Ans(*self.eval_components(*args, **kwargs))
+
 
+class BaseGradientModel(BaseCallableModel):
+    """
+    Baseclass for models which have a gradient. Such models are expected to
+    implement an `eval_jacobian` function.
+
+    Any subclass of this baseclass which does not implement its own
+    `eval_jacobian` will inherit a finite difference gradient.
```
pckroon: addition: that should *not* be used during the fit.

symfit/core/fit.py
```diff
@@ -408,7 +578,8 @@ def eval_jacobian(self, *args, **kwargs):
         """
         :return: The jacobian matrix of the function.
         """
-        return self.finite_difference(*args, **kwargs)
+        Ans = variabletuple('Ans', self)
+        return Ans(*self.finite_difference(*args, **kwargs))
```
pckroon: `BaseHessianModel`?

symfit/core/fit.py
```diff
@@ -488,148 +662,85 @@ def jacobian(self):
...
         jac = []
-        for param in self.params:
-            # Differentiate to every param
-            f = sympy.diff(self.chi, param)
-            jac.append(f)
+        for var, expr in self.items():
+            jac.append([])
+            for param in self.params:
+                partial_dv = D(var, param)
+                jac[-1].append(self.jacobian_model[partial_dv])
```
pckroon: I would build the new list first, and append it at the end instead of vice-versa

symfit/core/fit.py
```diff
@@ -488,148 +662,85 @@ def jacobian(self):
...
-    @cached_property
-    def numerical_hessian(self):
-        """
-        :return: lambda functions of the Hessian matrix of the function, which
-        can be used in numerical optimization.
-        """
-        return [[[sympy_to_py(second_order_pdv, self.independent_vars, self.params)
-                    for second_order_pdv in row]
-                for row in comp]
-            for comp in self.hessian]
```
pckroon: `HessianModel`?

symfit/core/fit.py
```diff
@@ -488,148 +662,85 @@ def jacobian(self):
...
 
-    @cached_property
-    def numerical_chi_squared(self):
-        """
-        :return: lambda function of the ``.chi_squared`` method, to be used in
-            numerical optimisation.
-        """
-        return sympy_to_py(self.chi_squared, self.vars, self.params)
+class Model(GradientModel):
```
pckroon: Should inherit from `HessianModel`?

symfit/core/fit.py
```diff
@@ -488,148 +662,85 @@ def jacobian(self):
...
-        Vector of derivatives w.r.t. each parameter. Not a Matrix but a vector!
+        :return: Jacobian evaluated at the specified point.
         """
-        jac = []
-        for param in self.params:
-            # Differentiate to every param
-            f = sympy.diff(self.chi_squared, param)
-            jac.append(f)
-        return jac
+        eval_jac_dict = self.jacobian_model(*args, **kwargs)._asdict()
```
pckroon: Who deals with the topological sorting here?

symfit/core/fit.py
```diff
@@ -488,148 +662,85 @@ def jacobian(self):
...
         """
         :return: Hessian evaluated at the specified point.
         """
-        hess = [[[np.atleast_1d(second_order_pdv(*args, **kwargs))
-                    for second_order_pdv in row]
-                for row in comp]
-            for comp in self.numerical_hessian
+        # Evaluate the hessian model and use the resulting Ans namedtuple as a
+        # dict. From this, take the relevant components.
+        eval_hess_dict = self.hessian_model(*args, **kwargs)._asdict()
```
pckroon: Who deals with the topological sorting here?

symfit/core/fit.py
```diff
@@ -719,23 +844,35 @@ class Constraint(Model):
...
+            of the corresponiding model, either by providing the model or the
+            parameters.
+        :param params: list of parameters in case no ``model`` is provided.
         """
+        params = kwargs.pop('params')
         if isinstance(constraint, Relational):
             self.constraint_type = type(constraint)
-            if isinstance(model, BaseModel):
+            if model is None and params is not None:
+                pass
```
pckroon: Put `self.params = params` here

symfit/core/fit.py
```diff
@@ -719,23 +844,35 @@ class Constraint(Model):
...
+            elif isinstance(model, BaseModel):
                 self.model = model
+                params = self.model.params
```
pckroon: And `self.params = self.model.params` here

symfit/core/fit.py
```diff
@@ -719,23 +844,35 @@ class Constraint(Model):
...
             else:
                 raise TypeError('The model argument must be of type Model.')
             super(Constraint, self).__init__(constraint.lhs - constraint.rhs)
 
             # Update the signature to accept all vars and parms of the model
-            self.independent_vars = self.model.vars
-            self.params = self.model.params
+            self.params = params
```
pckroon: And finally remove this

symfit/core/fit.py
```diff
@@ -1062,11 +1179,12 @@ def _cov_mat_equal_lenghts(self, best_fit_params):
...
-        jac = np.array(self.model.eval_jacobian(**kwargs))
+        jac = self.model.eval_jacobian(**kwargs)
         # Drop the axis which correspond to dependent vars which have been
-        # set to None
-        mask = [data is not None for data in self.dependent_data.values()]
-        jac = jac[mask]
+        # set to None. jac also contains interdependent_vars, hence the .get
+        mask = [self.dependent_data.get(var, None) is not None
+                for var in self.model.dependent_vars]
+        jac = np.array([comp for comp, relevant in zip(jac, mask) if relevant])
```
pckroon: Comment: can't do numpy indexing with mask because `jac` is a list.

symfit/core/fit.py
```diff
@@ -1237,18 +1346,14 @@ def covariance_matrix(self, best_fit_params):
...
         kwargs = {p.name: float(value) for p, value in best_fit_params.items()}
         kwargs.update(self.independent_data)
-        # kwargs.update(self.data)
-        X = np.atleast_2d([
-            (np.ones(sigma.shape[0]) * comp(**key2str(kwargs))).flatten()
-            for comp in self.model.numerical_jacobian[0]
-        ])
-
+        X = np.atleast_2d(self.model.eval_jacobian(**key2str(kwargs))[0])
+        X = np.ones(sigma.shape[0]) * X
```
pckroon: `np.broadcast_to`

symfit/core/fit.py
```diff
@@ -1803,7 +1692,7 @@ def r_squared(model, fit_result, data):
     SS_tot = np.sum([np.sum((y_i - y_bar)**2) for y_i, y_bar in zip(y_is, y_bars) if y_i is not None])
     return 1 - SS_res/SS_tot
 
-class ODEModel(CallableModel):
+class ODEModel(BaseGradientModel):
```
pckroon: *not* a gradient model!

symfit/core/fit.py
```diff
@@ -1832,6 +1721,7 @@ def __init__(self, model_dict, initial, *lsoda_args, **lsoda_kwargs):
             key=sort_func
         )
         self.independent_vars = sorted(set(d.variables[0] for d in model_dict), key=sort_func)
+        self.interdependent_vars = []  # Not yet supported for ODEModels
```
pckroon: Why? What's the problem?

symfit/core/fit.py
```diff
@@ -2029,7 +1919,108 @@ def __call__(self, *args, **kwargs):
...
+        Sympy does not handle repeated partial derivation correctly, e.g.
+        D(D(y, a), a) = D(y, a, a) but D(D(y, a), b) = 0.
+        Use this function instead to prevent evaluation to zero.
+        """
+        if isinstance(var, sympy.Derivative):
+            return sympy.Derivative(var.expr, *(var.variables + params))
+        else:
+            return D(var, *params)
+
+    def partial_subs(func, func2vars):
```
pckroon: Move these two helper functions outside, and prefix them with a `_`

symfit/core/fit.py
```diff
@@ -2029,7 +1919,108 @@ def __call__(self, *args, **kwargs):
...
+
+    This function make sure the chain rule is correctly applied for
+    interdependent variables.
+
+    :param model: Any symbolical model-type.
+    :return: :class:`~symfit.core.fit.CallableModel` representing the Hessian
+        of ``model``.
+    """
+    jac_model = jacobian_from_model(model, as_functions=True)
+    return jacobian_from_model(jac_model)
```
pckroon: Smooth.

symfit/core/fit.py
```diff
@@ -2029,7 +1919,108 @@ def __call__(self, *args, **kwargs):
...
+        of ``model``.
+    """
+    chi2_expr = sum(((model[y] - y)**2 / model.sigmas[y]**2)
+                    for y in model.dependent_vars)
+    chi2 = Variable()
+    chi2_dict = {chi2: chi2_expr}
+    # Update it with any leftover interdependent variables we might need.
+    chi2_dict.update({var: expr for var, expr in model.items()
+                     if var not in model.dependent_vars})
+    return GradientModel(chi2_dict)
```
pckroon: Why no HessianModel? We know it's going to be 0, but we might as well tell whoever is going to use this in the future.

symfit/core/printing.py
```diff
@@ -31,6 +31,32 @@ class SymfitNumPyPrinter(NumPyPrinter):
...
+        result = [printouts[0]]
+        for link, printout in zip(links, printouts[1:]):
+            result.extend([link, printout])
+        return '({0})'.format(''.join(result))
+
+    def _print_Inverse(self, expr):
+        return "%s(%s)" % (self._module_format('numpy.linalg.inv'),
+                           self._print(expr.args[0]))
+
+    def _print_HadamardProduct(self, expr):
```
pckroon: I'm guessing this is the element wise product?

symfit/core/support.py
```diff
@@ -357,8 +375,18 @@ def wrapped_func(*args, **kwargs):
         return wrapped_func
 
 
-class D(sympy.Derivative):
+def D(*args, **kwargs):
+    return sympy.Derivative(*args, **kwargs)
```
pckroon: Why the change?

symfit/core/support.py
```diff
@@ -357,8 +375,18 @@ def wrapped_func(*args, **kwargs):
...
+    Save name which can be used for alphabetic sorting and can be turned
+    into a kwarg.
     """
+    base_str = 'd{}{}_'.format(self.derivative_count if
+                               self.derivative_count > 1 else '', self.expr)
+    for var, count in self.variable_count:
+        base_str += 'd{}{}'.format(var,  count if count > 1 else '')
+    return base_str
+
+sympy.Derivative.name = property(name)
```
pckroon: Ew.

tests/test_model.py
```diff
@@ -219,9 +225,169 @@ def test_pickle(self):
             # Compare signatures
             self.assertEqual(model.__signature__, new_model.__signature__)
             # Trigger the cached vars.
-            model.vars
-            new_model.vars
-            self.assertEqual(new_model.__dict__, model.__dict__)
+            model.vars, model.connectivity_mapping
+            new_model.vars, new_model.connectivity_mapping
```
pckroon: Use 4 lines.

tests/test_model.py
```diff
@@ -219,9 +225,169 @@ def test_pickle(self):
...
+            if not isinstance(model, ODEModel):
+                model.function_dict, model.vars_as_functions
+                new_model.function_dict, new_model.vars_as_functions
```
pckroon: Here as well

tests/test_model.py
```diff
@@ -219,9 +225,169 @@ def test_pickle(self):
...
+        self.assertEqual(hess_model.interdependent_vars,
+                         [D(y, (a, 2)), D(y, a), D(y, b), y])
+        self.assertEqual(hess_model.independent_vars, [x])
+
+
+        model = Model(model_dict)
+        np.testing.assert_almost_equal(model(x=3, a=1, b=2),
+                                       np.atleast_2d([7, 51]).T)
+        np.testing.assert_almost_equal(model.eval_jacobian(x=3, a=1, b=2),
+                                       np.atleast_2d([[[9], [4]], [[128], [57]]]))
```
pckroon: Shouldn't this be at least 3d?

tests/test_model.py
```diff
@@ -219,9 +225,169 @@ def test_pickle(self):
...
+        np.testing.assert_almost_equal(
+            model.eval_hessian(x=3, a=1, b=2),
+            np.atleast_2d([[[[18], [0]], [[0], [2]]],
```
pckroon: Shouldn't this be at least 4d? (I hope there's a function for that...)

pckroon:
Referencing #39

symfit/core/fit.py
```diff
@@ -133,23 +170,98 @@ def _init_from_dict(self, model_dict):
 
         :param model_dict: dict of (dependent_var, expression) pairs.
         """
-        sort_func = lambda symbol: str(symbol)
-        self.model_dict = OrderedDict(sorted(model_dict.items(), key=lambda i: sort_func(i[0])))
-        self.dependent_vars = sorted(model_dict.keys(), key=sort_func)
+        sort_func = lambda symbol: symbol.name
+        self.model_dict = OrderedDict(sorted(model_dict.items(),
+                                             key=lambda i: sort_func(i[0])))
```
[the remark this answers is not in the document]
tBuLi: We iterate over the items, which are two-tuples. So we sort by the first element of the tuple.

symfit/core/fit.py
```diff
@@ -133,23 +170,98 @@ def _init_from_dict(self, model_dict):
...
+        # Everything at the bottom of the toposort is independent, at the top
+        # dependent, and the rest interdependent.
+        ordered = list(toposort(self.connectivity_mapping))
+        independent = ordered.pop(0)
```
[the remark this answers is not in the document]
tBuLi: good catch

symfit/core/fit.py
```diff
@@ -133,23 +170,98 @@ def _init_from_dict(self, model_dict):
...
+        self.dependent_vars = sorted(ordered.pop(-1), key=sort_func)
+        self.interdependent_vars = sorted(
+            [item for items in ordered for item in items],
+            key=sort_func
+        )
+        # `independent` contains both params and vars, needs to be separated
+        self.params = sorted(
+            [s for s in independent if isinstance(s, Parameter)],
```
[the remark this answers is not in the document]
tBuLi: Has been abandoned for now :P. It is not useful when it comes to sorting out the interdependence because it only deals with one `Expr` at a time, and since the Parameters are already present at the lowest level of the toposort it makes more sense to just take them from there.

symfit/core/fit.py
tBuLi: I'll add some comments to the code ;). (And do some renaming of variables) Basically, `functions` will map vars to functions, right? And since we walk the tolopological tree in order, I substitute every variable in connections by its corresponding function if it is already present in `function`. Due to the order, this is guaranteed to work. So e.g. if I have
```python
model_dict = {z: y, y: x}
```
then the order will be `[x, y, z]`. Then, since `x` is not a key in connectivity mapping, it gets ignored. For `y`, we add `{y: Function('y')(x)}` to `functions`. For `z`, we add `{z: Function('z')(Function('y')(x))}`, which upon printing looks a lot more readable:
```python
{y: y(x), z: z(y(x))}
```

pckroon: Ah ok. I think it finally did "click". Still requires more comments/variable names though.

symfit/core/fit.py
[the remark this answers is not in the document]
pckroon: IIRC that function is still in the code and used, but fair enough.

symfit/core/fit.py
tBuLi: That would print every component as a function of all vars and params, but I think it is nicer in light of the new call signatures, where from now on we will only call expressions with the kwargs they actually need, that it is also better to print them with the vars they need.

E.g. `{y: a * x, z: x}` prints as
```python
[y(x; a) = a * x,
 z(x) = x]
```
It might also be nice to directly print from function dict instead, giving
```python
{y(x, a): a * x,
 z(x): x}
```
This might be nicer because it is also essentially valid code, but what I like about the current implementation is that it is more mathematical.

pckroon: Make that the `__repr__` together with a `Model` around it.

symfit/core/fit.py
tBuLi: Because I copy-pasted this from basemodel. It might be better to start raising these warnings.

tBuLi: Also in BaseModel, to make users think about it.

symfit/core/fit.py
tBuLi: I don't think that we can if we don't want to use inspect to much. Without using `inspect` on the provided functions we cannot know which kwargs to provide.

We could perhaps check for the existence of an `__signature__` attribute, but that is not present on most callable by default to my knowledge, unless explicitly set.

The good news is, that making everything a function of everything was the convention up to this point, so we now just do better if we can.

symfit/core/fit.py
tBuLi: Good point.

symfit/core/fit.py
tBuLi: Then what is it there for? :P This finite difference jacobian is used during the fit, it comes in particularly handy for ODEModels which otherwise would not have a gradient (because I haven't had time/the need to go through the mathematics.)

pckroon: It's currently *not* used for the fit itself, just for the calculation/estimation of the covariance matrix. It's too expensive to calculate it every minimisation step --- doing the minimisation without or with approximated jacobians is faster.

symfit/core/fit.py
tBuLi: The jacobian_model takes care of the sorting internally, being a fully self-dependent `CallableModel`.

symfit/core/fit.py
tBuLi: Not sure I understand this remark, you mean why does it not inherit from GradientModel? Because for ODEModels it is hard to estimate the jacobian analytically, since that will itself be a system of ODEs. It's much easier to inherit the finite differences from BaseGradientModel.

symfit/core/fit.py
tBuLi: Mostly that I haven't thought about it yet ;). ODEModels are a pain in the ass.

I think actually that after this restructuring, ODEModels might be reinvented as some kind of  NumericalModel. Because that is mostly what they are. But more though on this is needed.

symfit/core/printing.py
tBuLi: Yeah, that's the formal mathematical name for it, and how it's called in sympy.

symfit/core/support.py
tBuLi: In essence, I wanted to give these derivatives a name property. But because of the way sympy works, the `__new__` method creates a new `Expr`, whatever you do. So the simple inheritance doesn't work. In the old way of doing `class D(sympy.Derivative)`, with a name property on it, the resulting object is still a `sympy.Derivative`, not a `D`. 

Maybe I should spend some more time trying to understand exactly how this works to see if we can change it, but to get on with the rest of the code development I choose this solution.

symfit/core/fit.py
pckroon: No, vice versa. `ODEModel` should *not* inherit from `BaseGradientModel` *because* we don't have an analytical Jacobian. And finite difference is too expensive to use during the minimization.

symfit/core/fit.py
pckroon: Add a TODO or FIXME comment ;)

symfit/core/support.py
pckroon: I was afraid of something like that. Put a TODO or FIXME comment here for now then.

symfit/core/fit.py
pckroon: So why didn't you?
<Palpatine_saying_do_it.gif>

symfit/core/fit.py
pckroon: However, we can estimate the jacobian using finite differences for the calculation of the covariance matrix

symfit/core/fit.py
tBuLi: Currently it is being used actually. (tested with a simple print) Because if a model has an eval_jacobian, that jacobian is passed on to the minimizer. Or rather, if both the model and the objective have eval_jacobian.

This is indeed new behavior, because before we checked if the model had `numerical_jacobian`. Perhaps we can now change that to checking for `jacobian_model` on the model, or accept that if a model inherits directly from BaseGradientModel, currently only ODEModel, it has a more expensive jacobian.

Because I do think the current decision tree makes more sense than before.

pckroon: It shouldn't ;)
Maybe move finite difference to `BaseModel` or `CallableModel` since we can do it with *any* model? And then check if it's a `JacobianModel` to see if we pass anything to the minimizer/objective.

symfit/core/minimizers.py
```diff
@@ -60,11 +60,13 @@ def _baseobjective_from_callable(self, func, objective_type=MinimizeModel):
...
                 model = func
             else:
-                # Minimize the provided custom objective instead. This why want to
-                # minimize a CallableNumericalModel, thats what they are for.
-                model = CallableNumericalModel(func,
-                                               params=self.parameters,
-                                               independent_vars=[])
+                # Minimize the provided custom objective instead. We therefore
+                # wrap it into a CallableNumericalModel, thats what they are for
+                y = sympy.Dummy('y')
```
pckroon: How bad is this going to go if the model uses a variable `'y'` somewhere? I guess it should be OK since it's a Dummy, right?
Does it work in general with all the sorting into variables/parameters if users make a model with both Dummies and Variables with identical names? That question is out of scope though.

tests/test_model.py
```diff
@@ -218,18 +218,29 @@ def test_pickle(self):
...
+                  connected_num_model]
+        for model in models:
             new_model = pickle.loads(pickle.dumps(model))
             # Compare signatures
             self.assertEqual(model.__signature__, new_model.__signature__)
             # Trigger the cached vars.
-            model.vars, model.connectivity_mapping
-            new_model.vars, new_model.connectivity_mapping
+            model.vars
+            new_model.vars
```
pckroon: Add a comment that we trigger the cached attributes *because we compare `__dict__` s*

tBuLi:
Don't worry, it's still very much a work in progress ;). The non-trivial comments I left open were things I wanted to think about carefully before implementing them.

For future reference: this mentiones how the hessian relates to the covariance matrix:  https://www8.cs.umu.se/kurser/5DA001/HT07/lectures/lsq-handouts.pdf

My suggestion is to remove finite differences entirely, and to always use the Hessian of the objective to calculate covariances. This removes the discussion about using finite differences for optimization and will finally clean-up the mess that is covariance matrix calculation at this point.

pckroon:
> My suggestion is to remove finite differences entirely, and to always use the Hessian of the objective to calculate covariances. This removes the discussion about using finite differences for optimization and will finally clean-up the mess that is covariance matrix calculation at this point.

Bad plan, we still need it to get a covariance matrix for e.g. ODEModels and NumericalModels. What we can do, however, is deprecate the finite difference Jacobian, and change it for a finite difference Hessian method. Or keep both next to each other.

symfit/core/fit.py
pckroon: With https://github.com/tBuLi/symfit/pull/201/commits/e173162581cf9716f01a3105c545914e5caacca7 this is now also dead weight, correct?

symfit/core/objectives.py
```diff
@@ -322,7 +322,7 @@ def eval_hessian(self, ordered_parameters=[], **parameters):
                 p1 = hess_comp * ((y - f) / sigma**2)[np.newaxis, np.newaxis, ...]
                 # Outer product
                 p2 = np.einsum('i...,j...->ij...', jac_comp, jac_comp)
-                p2 /= sigma[np.newaxis, np.newaxis, ...]**2
+                p2 = p2 / sigma[np.newaxis, np.newaxis, ...]**2
```
pckroon: Why not `/=`?

symfit/core/fit.py
```diff
@@ -1009,169 +1010,78 @@ class HasCovarianceMatrix(TakesData):
...
-                    degrees_of_freedom += np.product(data.shape)
+        if isinstance(objective, LogLikelihood):
+            # Loglikelihood is a special case that needs to be considered
+            # separately, see #138.
+            return hess_inv
+        elif isinstance(objective, LeastSquares):
+            # Calculate the covariance for a least squares method.
+            # https://www8.cs.umu.se/kurser/5DA001/HT07/lectures/lsq-handouts.pdf
+            # Residual sum of squares
+            rss = objective(**key2str(best_fit_params))
```
pckroon: The RSS should probably be a "goodness of fit" descriptor, and every minimizer will return it. Would save a function eval

symfit/core/fit.py
```diff
@@ -1009,169 +1010,78 @@ class HasCovarianceMatrix(TakesData):
...
-                else: # the correspoding component in ss_res will be 0 so it is ok to add any non-zero number.
-                    degrees_of_freedom.append(len(best_fit_params) + 1)
-        degrees_of_freedom = np.array(degrees_of_freedom)
-        s_sq = ss_res / (degrees_of_freedom - len(best_fit_params))
-        # s_sq = ss_res / degrees_of_freedom
-
-        return s_sq
+                s2 = rss / dof
+            # Divide by two because the source uses different normalization
+            cov_mat = 2 * s2 * hess_inv
```
pckroon: Divide?

symfit/core/fit.py
```diff
@@ -1009,169 +1010,78 @@ class HasCovarianceMatrix(TakesData):
...
+            return cov_mat
+        else:
+            # We do not know how to handle with this situation.
+            return None
```
pckroon: This If/elif totem with isinstance calls probably means objectives should take care of calculating the covariance matrix

symfit/core/fit.py
```diff
@@ -1009,169 +1010,78 @@ class HasCovarianceMatrix(TakesData):
...
         """
-        # Stack in a new dimension, and make this the first dim upon indexing.
-        sigma = np.concatenate([arr[np.newaxis, ...] for arr in self.sigma_data.values()], axis=0)
+        cov_matrix = self._covariance_matrix(best_fit_params,
+                                             objective=self.objective)
+        if cov_matrix is None:
+            # If the covariance matrix could not be computed, we try again by
+            # approximating the hessian with the jacobian.
+            if not isinstance(self.objective, HessianObjective) \
+                    or not isinstance(self.model, HessianModel):
```
pckroon: Why this if? We already know we couldn't make a cov_matrix using hessians

symfit/core/objectives.py
```diff
@@ -329,6 +329,39 @@ def eval_hessian(self, ordered_parameters=[], **parameters):
         return np.atleast_2d(np.squeeze(np.array(result)))
 
 
+class HessianObjectiveJacApprox(HessianObjective):
+    """
+    This object should only be used as a Mixin for covariance matrix estimation.
```
pckroon: Would it not be better to make this a part of `JacobianObjective`? That way there always is a sane way of evalling a Hessian. It would also obliviate the need for `HessApproximation` above.

symfit/core/fit.py
```diff
@@ -1057,30 +1057,30 @@ def covariance_matrix(self, best_fit_params):
...
-                    """
-
-                objective = HessApproximation(self.objective.model,
-                                              self.objective.data)
-                cov_matrix = self._covariance_matrix(best_fit_params,
-                                                     objective=objective)
+        if cov_matrix is None or any(np.diag(cov_matrix) < 0):
+            # If the covariance matrix could not be computed or contains illegal
+            # values, we try again by approximating the hessian with the
+            # jacobian.
```
pckroon: More importantly, how/why/when does this happen? And how negative are those values?

tests/test_constrained.py
```diff
@@ -489,9 +519,17 @@ def test_error_advanced(self):
...
+        # Hessian we actually get a more accurate value from the standard fit
+        # then for MINPACK. Hence we check if it is roughly equal, and if our
+        # stdev is greater than that of minpack.
+        self.assertAlmostEqual(const_result.stdev(a) / std_result.stdev(a), 1, 2)
+        self.assertAlmostEqual(const_result.stdev(b) / std_result.stdev(b), 1, 1)
+        self.assertAlmostEqual(const_result.stdev(c) / std_result.stdev(c), 1, 2)
+
+        self.assertGreaterEqual(const_result.stdev(a), std_result.stdev(a))
+        self.assertGreaterEqual(const_result.stdev(b), std_result.stdev(b))
+        self.assertGreaterEqual(const_result.stdev(c), std_result.stdev(c))
```
pckroon: Does this mean MINPACK *under*estimates covariances?

tests/test_general.py
```diff
@@ -707,34 +706,38 @@ def test_error_advanced(self):
         z = Variable('z')
         model = {z: a * log(b * x + c * y)}
 
-        # fit = Fit(model, xy, z, absolute_sigma=False)
-        fit = Fit(model, xdata, ydata, zdata, absolute_sigma=False)
-        # fit = Fit(model, x=xdata, y=ydata, z=zdata, absolute_sigma=False)
+        # Use a gradient model because Mathematica uses the Hessian
+        # approximation instead of the exact Hessian.
```
pckroon: That's interesting, since Mathematica does have access to the Hessian... Using the Hessian is supposed to be more accurate, right?

tests/test_objectives.py
```diff
@@ -90,7 +90,7 @@ def test_LeastSquares(self):
                          (2, 100))
         # Hessian no longer depends on the data, so its a scalar in the last dim
         self.assertEqual(model.eval_hessian(x=xdata, a=2, b=3)[0].shape,
-                         (2, 2, 1))
+                         (2, 2, 100))
```
pckroon: Comment is out of date?

tests/test_constrained.py
```diff
@@ -290,10 +290,17 @@ def test_vector_parameter_error(self):
...
-        As it stands now, `Fit` is able to correctly
-        predict the values of the parameters an their standard deviations, but
-        it is not able to give the covariances. Those are therefore returned as
-        nan, to prevent people from citing them as 0.0.
+        As per 0.5.0 this test has been updated in an important way. Previously
+        the covariance matrix was estimated on a per component basis for global
+        fitting problems. This was incorrect, but no solution was possible at
+        the time. Now, we calculate the covariance matrix from the Hessian of
+        the function being optimized, and so now the covariance is calculated
+        correctly in those scenarios.
```
pckroon: Sweet.

symfit/core/minimizers.py
tBuLi: Even `Dummy('y') == Dummy('y')` is `False`, the name is not important, essentially just the memory address.

Good question about the sorting, in principle they do have a name so I guess they would rank as the same for alphabetical ordering. That could actually go horribly wrong when both of those happen to be in the models signature. (Not a problem here, although I will nonetheless remove the name and let sympy generate one for us just in case.)

symfit/core/objectives.py
tBuLi: Broadcasting issues if the shape is changed during this operation.

symfit/core/fit.py
```diff
@@ -465,6 +467,14 @@ def numerical_components(self):
              '`numerical_components`.').format(self.__class__)
         )
 
+    def _get_params(self):
+        return self._params
+
+    def _set_params(self, value):
+        self._params = value
+        self.__signature__ = self._make_signature()
+    params = property(_get_params, _set_params)
```
pckroon: Why not use the `@property` decorator?

symfit/core/fit.py
```diff
@@ -648,6 +658,12 @@ def __init__(self, *args, **kwargs):
         super(GradientModel, self).__init__(*args, **kwargs)
         self.jacobian_model = jacobian_from_model(self)
 
+    def _set_params(self, value):
+        super(GradientModel, self)._set_params(value)
+        if hasattr(self, 'jacobian_model'):
+            self.jacobian_model.params = value
+    params = property(CallableModel._get_params, _set_params)
```
pckroon: As decorator?

symfit/core/fit.py
```diff
@@ -694,6 +710,12 @@ def __init__(self, *args, **kwargs):
         super(HessianModel, self).__init__(*args, **kwargs)
         self.hessian_model = hessian_from_model(self)
 
+    def _set_params(self, value):
+        super(HessianModel, self)._set_params(value)
+        if hasattr(self, 'hessian_model'):
+            self.hessian_model.params = value
+    params = property(GradientModel._get_params, _set_params)
```
pckroon: As decorator?

symfit/core/fit.py
```diff
@@ -1143,10 +1120,22 @@ def __init__(self, model, *ordered_data, **named_data):
...
+                                                  model=model)
         super(Fit, self).__init__(model, *ordered_data, **named_data)
 
-        # List of Constraint objects
-        self.constraints = self._init_constraints(constraints=constraints)
+        # Update the data belonging to the constraints
+        for constraint in self.constraints:
+            for var in constraint.independent_vars:
+                if var.name in self.data:
+                    self.data[var] = self.data.pop(var.name)
```
pckroon: This needs a comment and/or better initialization so it's not needed.

symfit/core/fit.py
```diff
@@ -1185,6 +1174,27 @@ def __init__(self, model, *ordered_data, **named_data):
         else:
             self.minimizer = self._init_minimizer(minimizer)
 
+    def _make_signature(self):
+        signature = super(Fit, self)._make_signature()
+
+        var_names = [var.name for constraint in self.constraints
+                     for var in constraint.independent_vars]
+        var_names = set(var_names)
```
pckroon: Can you do a set comprehension in Py2? `var_names = {var.name for ...}`

symfit/core/fit.py
```diff
@@ -1185,6 +1174,27 @@ def __init__(self, model, *ordered_data, **named_data):
...
+        # Extra parameter objects due to vars of constraints.
+        extra_parameters = []
+        for var_name in var_names:
+            if var_name not in signature.parameters:
```
pckroon: Comment: Not symfit Parameter

symfit/core/fit.py
```diff
@@ -1626,4 +1623,81 @@ def hessian_from_model(model):
...
+            # work the model into the constraint
+            model_dict = constraint_model.model_dict.copy()
+            constraint_type = constraint_model.constraint_type
+            for var in constraint_model.independent_vars:
+                if var in model:
+                    model_dict[var] = model[var]
+                    # Also add possible dependencies of this var
+                    for symbol in model.ordered_symbols:
+                        # Walk over all possible dependencies of this
+                        # variable until we no longer have dependencies.
```
pckroon: Does this also deal with dependencies of dependencies of dependencies ...?

symfit/core/minimizers.py
```diff
@@ -405,12 +405,15 @@ def scipy_constraints(self, constraints):
...
-                constraint = MinimizeModel(constraint, data={})
+                constraint = MinimizeModel(constraint, data=self.objective.data)
             elif isinstance(constraint, sympy.Rel):
-                from .fit import Constraint, CallableNumericalModel
-                constraint = Constraint(constraint, params=self.parameters)
+                from .fit import prepare_constraint
+                constraint = prepare_constraint(
+                    constraint, self.parameters,
+                    model_type=self.objective.model.__class__
+                )
```
pckroon: I still don't think this constraint init code should be here. If you want to use Minimizers directly, you need to provide the correct arguments. I do agree that the API of Fit, Model, etc should be as flexible and user friendly as possible.

symfit/core/minimizers.py
```diff
@@ -460,9 +463,12 @@ def scipy_constraints(self, constraints):
         # Take the normal scipy compatible constraints, and add jacobians.
         scipy_constr = super(SLSQP, self).scipy_constraints(constraints)
         for scipy_constraint in scipy_constr:
-            scipy_constraint['jac'] = self.resize_jac(
-                scipy_constraint['fun'].eval_jacobian
-            )
+            # Only if the model has a jacobian, does it make sense to pass one
+            # to the minimizer
+            if hasattr(scipy_constraint['fun'].model, 'eval_jacobian'):
```
pckroon: I don't know. Does it hurt to just always pass a constraint jac if we have one?

symfit/core/fit.py
```diff
@@ -133,23 +166,96 @@ def _init_from_dict(self, model_dict):
...
+        # dependent, and the rest interdependent.
+        ordered = list(toposort(self.connectivity_mapping))
+        independent = sorted(ordered.pop(0), key=sort_func)
+        self.dependent_vars = sorted(ordered.pop(-1), key=sort_func)
+        self.interdependent_vars = sorted(
+            [item for items in ordered for item in items],
+            key=sort_func
+        )
+        # `independent` contains both params and vars, needs to be separated
+        self.params = [s for s in independent if isinstance(s, Parameter)]
```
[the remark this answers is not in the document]
tBuLi: Excellent idea ;)

symfit/core/fit.py
tBuLi: Unfortunately you cannot call super from within a `property` :(. This means that a construction like the one above is the only way to avoid code reuse. It looks dreadful internally I know, but at least from the outside API this is equivalent.

symfit/core/fit.py
tBuLi: I agree that it is weird, we are not talking about small value here. It happens for `TestFitResults.test_fitting_2`, and the variances can be as big as `-1150`. Since this means an imaginary standard deviation, the quickest fix is to do the Jacobian approximation, which is guaranteed to be positive-definite.

I'm not sure why it happens, maybe it's something like this: https://stats.stackexchange.com/questions/30465/what-does-a-non-positive-definite-covariance-matrix-tell-me-about-my-data.

Maybe this test is poorly posed to begin with, though it is not an uncommon situation.

tBuLi: p.s. I've now made all the values in this test less ideal so the components become independent, but the problem persists.

symfit/core/fit.py
tBuLi: Yes, due to the ordering everything is passed on because we start with the low hanging fruit and work our way up

symfit/core/minimizers.py
tBuLi: Yes, I kinda agree, but I experienced in my own codebase that this is a very nice feature to have, and not that much work for us.

symfit/core/fit.py
tBuLi: I agree, but maybe we should keep that for when we fix the goodness of fit mayhem ;).

tests/test_constrained.py
tBuLi: In this particular example, yes. But actually I should be a bit careful with my statements here, because I don't think it has to be a greater than in general, just that it is not the same.

tests/test_general.py
tBuLi: Yes. I think it might be a convention/computational ease thing?

symfit/core/fit.py
pckroon: Hmmnn.. To summarize the SE post: there are colinear Parameters? This should at the very least raise a *loud* warning to the user that something is fishy. We could even tell them which parameters are colinear; either automagically, or expose a function/method that tells them.

tests/test_general.py
pckroon: I've not caught Mathematica on either of those before. So I find it rather worrying.

symfit/core/fit.py
pckroon: Add a comment about this

symfit/core/minimizers.py
pckroon: > and not that much work for us.

I disagree there. I think it incurs a rather significant technical debt, but time will tell.

tBuLi: We can perhaps review this as a part of #203. I agree that perhaps the minimizer code is becoming bulky, which is definitely not supposed to happen.

symfit/core/objectives.py
tBuLi: I personally think it is weird for a `JacobianObjective` to implement a hessian. I also think it can lead to false positives elsewhere because we give the minimizer a Hessian if the model has `hessian_model` and the objective has `eval_hessian`. That could trigger a false positive here.

symfit/core/fit.py
tBuLi: Fit now checkes if a model has `jacobian_model` to determine if it should pass `objective.eval_jacobian` to the minimizer, and so anything inheriting from `BaseGradientModel` is now save in terms of the minimization process, but will still have a method to calculate the covariance matrix. Order has been restored to the universe.

symfit/core/fit.py
tBuLi: Yeah, maybe it is actually better to change our tests to accept NaN as the proper answer instead of trying to give the old result which is due to a (standard) approximation. The model used in the mentioned test does fail to include the covariance, so perhaps this is indeed due to a poor model. And [here ](https://www.researchgate.net/post/What_is_the_explanation_of_negative_variance_in_the_statistical_models)it is also mentioned that software should be honest about negative covariances, with which I completely agree. Let's just confront users with it so they can conclude for themselves what it means.

tBuLi: A better guess seems to have solved the problem, but I also stand by my previous comment.

symfit/core/minimizers.py
```diff
@@ -56,12 +56,15 @@ def _baseobjective_from_callable(self, func, objective_type=MinimizeModel):
...
         else:
-            # Minimize the provided custom objective instead. This why want to
-            # minimize a CallableNumericalModel, thats what they are for.
-            from .fit import CallableNumericalModel
-            model = CallableNumericalModel(func,
-                                           params=self.parameters,
-                                           independent_vars=[])
+            from .fit import CallableNumericalModel, BaseModel
+            if isinstance(func, BaseModel):
+                model = func
```
pckroon: https://github.com/tBuLi/symfit/issues/203#issuecomment-467062179

symfit/core/fit.py
```diff
@@ -356,32 +356,56 @@ class BaseNumericalModel(BaseModel):
...
+        :param connectivity_mapping: Mapping indicating the dependencies of
+            every variable in the model. For example, a model_dict
+            {y: a * x + b} has a connectivity_mapping {y: {x, a, b}}. Note that
+            the values of this dict have to be sets.
+        """
+        connectivity_mapping = kwargs.pop('connectivity_mapping')
+        if connectivity_mapping is None and \
+                independent_vars is not None and params is not None:
+            # Make model into a mapping if needed.
+            if not isinstance(model, Mapping):
```
pckroon: Maybe my complaint about what deals with user-input where is also valid here. If you want to use this directly, you should provide the correct object. Only `Fit` gets to be clever.

symfit/core/fit.py
```diff
@@ -2144,3 +2144,106 @@ def __call__(self, *args, **kwargs):
...
+
+    This function make sure the chain rule is correctly applied for
+    interdependent variables.
+
+    :param model: Any symbolical model-type.
+    :return: :class:`~symfit.core.fit.CallableModel` representing the Hessian
+        of ``model``.
+    """
+    jac_model = jacobian_from_model(model, as_functions=True)
+    print(jac_model)
```
pckroon: Remaining debug output

symfit/core/fit.py
```diff
@@ -827,23 +827,35 @@ class Constraint(Model):
...
         """
         :param constraint: constraint that model should be subjected to.
-        :param model: A constraint is always tied to a model.
+        :param model: A constraint needs to be initiated with all the parameters
+            of the corresponiding model, either by providing the model or the
+            parameters.
+        :param params: list of parameters in case no ``model`` is provided.
         """
+        params = kwargs.pop('params')
         if isinstance(constraint, Relational):
```
pckroon: Same complaint: who deals with user input?

symfit/core/fit.py
pckroon: By the way, turn ODEModels into NumericalModels in a different PR. Please. ;)

symfit/core/minimizers.py
```diff
@@ -406,13 +406,7 @@ def scipy_constraints(self, constraints):
                 constraint = MinimizeModel(constraint, data={})
             elif isinstance(constraint, sympy.Rel):
```
pckroon: Who deals with user input?

symfit/core/fit.py
```diff
@@ -670,7 +670,7 @@ def eval_jacobian(self, *args, **kwargs):
         """
         eval_jac_dict = self.jacobian_model(*args, **kwargs)._asdict()
         # Take zero for component which are not present, happens for Constraints
-        jac = [[eval_jac_dict.get(D(var, param), 0)
+        jac = [[eval_jac_dict.get(D(var, param), 0) * np.ones_like(eval_jac_dict[var])
```
pckroon: np.broadcast_to?

symfit/core/fit.py
```diff
@@ -709,7 +709,7 @@ def eval_hessian(self, *args, **kwargs):
         # Evaluate the hessian model and use the resulting Ans namedtuple as a
         # dict. From this, take the relevant components.
         eval_hess_dict = self.hessian_model(*args, **kwargs)._asdict()
-        hess = [[[eval_hess_dict.get(D(var, p1, p2), 0)
+        hess = [[[eval_hess_dict.get(D(var, p1, p2), 0) * np.ones_like(eval_hess_dict[var])
```
pckroon: np.broadcast_to?

symfit/core/fit.py
```diff
@@ -1009,169 +1010,78 @@ class HasCovarianceMatrix(TakesData):
...
-        :param best_fit_params: ``dict`` of best fit parameters as given by .best_fit_params()
-        :return: covariance matrix.
-        """
-        if not hasattr(self.model, 'eval_jacobian'):
+    def _covariance_matrix(self, best_fit_params, objective):
+        # Helper function for self.covariance_matrix.
+        try:
+            hess = objective.eval_hessian(**key2str(best_fit_params))
+        except AttributeError:
+            # Some models do not have an eval_jacobian, in which case we give up
```
pckroon: s/eval_jacobian/eval_hessian/

symfit/core/fit.py
```diff
@@ -1009,169 +1010,78 @@ class HasCovarianceMatrix(TakesData):
...
         :param best_fit_params: ``dict`` of best fit parameters as given by .best_fit_params()
+        :return: covariance matrix.
         """
-        # Stack in a new dimension, and make this the first dim upon indexing.
-        sigma = np.concatenate([arr[np.newaxis, ...] for arr in self.sigma_data.values()], axis=0)
+        cov_matrix = self._covariance_matrix(best_fit_params,
+                                             objective=self.objective)
+        if cov_matrix is None:
+            # If the covariance matrix could not be computed, we try again by
+            # approximating the hessian with the jacobian.
```
pckroon: This bit of code would become cleaner if there were 2 methods: `_cov_matrix_from_hessian` (is now `_covariance_matrix`, and `_cov_matrix_from_jacobian`

pckroon: That would also help in testing: is _cov_matrix_from_hessian ~= _cov_matrix_from_jacobian?

symfit/core/fit.py
```diff
@@ -1163,10 +1112,22 @@ def __init__(self, model, *ordered_data, **named_data):
         objective = named_data.pop('objective')
         minimizer = named_data.pop('minimizer')
         constraints = named_data.pop('constraints')
+        # List of Constraint objects
+        if not isinstance(model, BaseModel):
```
pckroon: Who deals with user input?

symfit/core/fit.py
```diff
@@ -1648,4 +1594,81 @@ def hessian_from_model(model):
...
+    elif constraint_type is sympy.Le:
+        # We change this to a Ge
+        constraint = relational.rhs - relational.lhs
+        constraint_type = sympy.Ge
+    else:
+        constraint = relational.lhs - relational.rhs
+
+    if not issubclass(model_type, BaseNumericalModel):
+        constraint_model = model_type(constraint)
+    else:
```
pckroon: Maybe give all candidates for this a `from_constraint` (or something) method? That would make this bit much cleaner.
See also: factory pattern, https://realpython.com/factory-method-python/

symfit/core/fit.py
```diff
@@ -1648,4 +1594,81 @@ def hessian_from_model(model):
...
+                    model_dict[var] = model[var]
+                    # Also add possible dependencies of this var
+                    for symbol in model.ordered_symbols:
+                        # Walk over all possible dependencies of this
+                        # variable until we no longer have dependencies.
+                        if symbol == var:
+                            break
+                        elif symbol in model:
+                            # Not everything in ordered_symbols is a key of model
+                            if symbol not in model_dict:
```
pckroon: `elif symbol in model and symbol not in model_dict:`

tests/test_constrained.py
```diff
@@ -615,5 +620,253 @@ def test_fixed_and_constrained(self):
...
+        fit = Fit(model, x=xdata, y=ydata, Y=2, constraints=[constraint])
+        fit = Fit(model, x=xdata, y=ydata)
+        fit = Fit(model, x=xdata, objective=LogLikelihood)
+
+        # Not allowed
+        with self.assertRaises(TypeError):
+            fit = Fit(model, x=xdata, y=ydata, Y=2)
+        with self.assertRaises(TypeError):
+            fit = Fit(model, x=xdata, y=ydata, Y=2, Z=3, constraints=[constraint])
+        # TODO: Uncomment these next lines when #214 has been fixed.
```
pckroon: It may be better to turn these into expected failure tests. But meh, work.

tests/test_constrained.py
```diff
@@ -615,5 +620,253 @@ def test_fixed_and_constrained(self):
...
+                con_map = fit.constraints[0].connectivity_mapping
+                self.assertEqual(con_map, {Y: {x, y}, y: {x, mu, sig, A}})
+
+            # Finally, test if the constraint worked
+            fit_result = fit.execute(options={'eps': 1e-15, 'ftol': 1e-10})
+            unconstr_value = fit.minimizer.wrapped_constraints[0]['fun'](**unconstr_result.params)
+            constr_value = fit.minimizer.wrapped_constraints[0]['fun'](**fit_result.params)
+            self.assertAlmostEqual(constr_value[0], 0.0, 10)
+            # And if it was very poorly met before
+            self.assertNotAlmostEqual(unconstr_value[0], 0.0, 2)
```
pckroon: unconstr tests should be moved outside of the loop

symfit/core/fit.py
```diff
@@ -191,6 +191,14 @@ def _init_from_dict(self, model_dict):
                                  not isinstance(s, Parameter) and not s in self]
         self.params = [s for s in independent if isinstance(s, Parameter)]
 
+        try:
+            assert not any(isinstance(var, Parameter)
+                           for var in self.dependent_vars)
+            assert not any(isinstance(var, Parameter)
+                           for var in self.interdependent_vars)
+        except AssertionError as err:
```
pckroon: This is not the way to do that.
```
if any(isinstance(var, Parameter) for van in self.dependent_vars) or any(isinstance(var, Parameter) for var in self.interdependent_vars):
    raise ModelError(...)
```

symfit/core/fit.py
```diff
@@ -1146,7 +1146,10 @@ def __init__(self, model, *ordered_data, **named_data):
                                                   model=model)
         super(Fit, self).__init__(model, *ordered_data, **named_data)
 
-        # Update the data belonging to the constraints
+        # Update the data belonging to the constraints. We do this by checking
+        # for the precense of data with the same name as one of the independent
```
pckroon: typo precense

symfit/core/minimizers.py
```diff
@@ -438,12 +452,16 @@ def execute(self, **minimize_options):
 
     def scipy_constraints(self, constraints):
         cons = super(ScipyHessianMinimize, self).scipy_constraints(constraints)
-        # FIXME: Just assume that because the objective has a hess, so do
-        # all the constraints
         for con in cons:
-            con['hess'] = self.resize_hess(con['fun'].eval_hessian)
+            # Only if the model has a jacobian, does it make sense to pass one
```
pckroon: s/jacobian/hessian

tests/test_fit_result.py
```diff
@@ -66,55 +66,68 @@ def test_fitting(self):
...
-        data_2 = np.random.multivariate_normal(mean, cov, 1000000)
-        data = np.vstack((data, data_2))
+        data_2 = np.random.multivariate_normal(mean, cov, 10**5)
+        data = np.vstack((data_1, data_2))
 
         # Insert them as y,x here as np fucks up cartesian conventions.
-        ydata, xedges, yedges = np.histogram2d(data[:,1], data[:,0], bins=200, range=[[0.0, 1.0], [0.0, 1.0]])
+        ydata, xedges, yedges = np.histogram2d(data[:, 1], data[:, 0], bins=200,
+                                               range=[[0.0, 1.0], [0.0, 1.0]],
+                                               density=True)
```
pckroon: `density` should probably be `normalize` or soemthing. Not looking it up now, I want to go home.

symfit/core/fit.py
tBuLi: You mean an alternative 'constructor'/initializer? That's a good idea.

tests/test_fit_result.py
tBuLi: No the opposite, according to their docs `density` is preferred over `normed`. So why it fails now on most python versions is a mystery to me.

pckroon: Add a `--upgrade` to the travis.yml, so that `pip install --only-binary numpy,scipy,matplotlib .[all]` becomes `pip install --upgrade --only-binary numpy,scipy,matplotlib .[all]`

tBuLi: doesn't seem to solve the problem

symfit/core/fit.py
```diff
@@ -105,10 +102,118 @@ def __init__(self, model, constraint_type=None):
...
+        # Check if the constraint_type is allowed, and flip the sign if needed
+        if constraint_type not in allowed_types:
+            raise ModelError(
+                'Only constraints of the type {} are allowed. A constraint'
+                ' of type {} was provided.'.format(allowed_types,
+                                                   constraint_type)
+            )
+        elif constraint_type is sympy.Le:
+            # We change this to a Ge and flip the sign
+            self = - self
```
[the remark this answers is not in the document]
tBuLi: Good remark, not anymore because the minus is applied to the whole model_dict, but from this PR onwards it should only apply to the dependent component of the model_dict. I'll update `__neg__`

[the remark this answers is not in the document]
pckroon: My problem is more conceptual:
```python3
class A:
    def __init__(self):
        self = 2

a = A()
print(a)
```
What is `a`?
Maybe doing `self *= -1` is more sane, since that is explicitely in-place (and possibly `__ineg__` instead of `__neg__`.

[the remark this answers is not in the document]
tBuLi: Ah, you mean that I'm messing with `self`? I see the confusion, but note that this is a classmethod right? So this is an alternative constructor which has to return an instance in the end. I already call that `self` because that is what it will end up becoming, but we can still manipulate it however we want at this stage.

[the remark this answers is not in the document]
tBuLi: And I we should keep supporting pure negation of models I think, because if you want to maximize something you can just say `Fit(- model, ...)` which is rather nice.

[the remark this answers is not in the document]
pckroon: Aaaah. I missed it's a class method. In that case change all `self` to `obj`. And yes, negation of models is an important feature.

PR #213, opened 2019-01-29
Coveralls: Coverage badge

tBuLi:
PR to add a coverage badge to symfit, coupled to travis builds.


pckroon:
If this also doesn't get the correct coverage try one of these. Pytest does run all the tests, so specifying where to look for the tests shouldn't help.
```
pytest --cov=symfit
pytest --cov=.
```

tBuLi:
When doing `--cov=.`, the result gets skewed because then test source code itself is also counted. My suspicion is that somehow my local pytest and travis are running with a somewhat different config

PR #218, opened 2019-03-06
Loglikelihood mutivariate

tBuLi:
After a recent [stackoverflow question](https://stackoverflow.com/questions/54967541/how-do-i-implement-maximum-likelihood-estimation-type-2), I decided to include a more elaborate example of likelihood fitting to the docs, which does this feature more justice.

### DIFF
--- a/docs/examples/ex_bivariate_likelihood.rst
+++ b/docs/examples/ex_bivariate_likelihood.rst
@@ -1,0 +1,22 @@
+Example: Likelihood fitting a Bivariate Gaussian
+================================================
+
+In this example, we shall perform likelihood fitting to a `bivariate normal
+distrubtion <http://mathworld.wolfram.com/BivariateNormalDistribution.html>`_,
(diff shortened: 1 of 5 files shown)


tBuLi:
Mostly it was just because this question made me realize that our API also works for multivariate loglikelihood problems nowadays, but so far we only had tests and examples of single variable problems. 

I agree with your second idea, I was actually thinking of extending this example by integrating out the `y` variable using sympy to get the marginal distribution, and see what happens then. But more elaborate examples of how sympy can make your life easier should definitely be included.

tBuLi:
I'll leave the marginal distribution for another time, it has some unforeseen complications because it is not a clean integral. It contains some mathematical object that cannot be transformed into numerical equivalents.

tBuLi:
It seems like this marginal distribution should in fact be a simple Gaussian again, but for some reason sympy does not fully simplify if no further hints are given. http://www.maths.qmul.ac.uk/~ig/MTH5118/Notes11-09.pdf

PR #221, opened 2019-03-10
FitResults cleanup

tBuLi:
Added the minimizer and objective to the `FitResults` object, since that is very relevant information. Also did some cleaning up of the tests, which mostly boiled down to removing a lot of redundant lines.


symfit/core/fit_results.py
```diff
@@ -17,32 +22,43 @@ class FitResults(object):
...
         """
         :param model: :class:`~symfit.core.fit.Model` that was fit to.
         :param popt: best fit parameters, same ordering as in model.params.
-        :param pcov: covariance matrix.
-        :param infodict: dict with fitting info.
-        :param mesg: Status message.
-        :param ier: Number of iterations.
+        :param covariance_matrix: covariance matrix.
+        :param minimizer_output: Raw output as given by the minimizer. This is
+            assumed to be a :class:`~scipy.optimize.OptimizationResult` instance.
```
pckroon: Now symfit becomes Scipy specific, barring the use of any other minimizer library. This seems undesirable.

symfit/core/fit_results.py
```diff
@@ -17,32 +22,43 @@ class FitResults(object):
...
+        self.params = OrderedDict(
+            [(p.name, value) for p, value in zip(self.model.params, popt)]
+        )
         self.covariance_matrix = covariance_matrix
+        self.gof_qualifiers = self._gof_qualifiers()
+
+    @cached_property
+    def iterations(self):
+        if hasattr(self.minimizer_output, 'nit'):
+            return self.minimizer_output.nit
```
pckroon: `return getattr(self.minimizer_output, 'nit', None)`, and I wouldn't cache it.

pckroon: Or any other property for FitResult, it just adds complexity for little to no gain.

symfit/core/fit_results.py
```diff
@@ -57,13 +73,13 @@ def __str__(self):
...
         res += 'Objective:              {}\n'.format(self.objective)
         res += 'Minimizer               {}\n'.format(self.minimizer)
-        try:
-            res += 'Regression Coefficient: {}\n'.format(self.r_squared)
-        except AttributeError:
-            pass
+
+        res += '\nGoodness of fit qualifiers:\n'
+        res += '\n'.join('{:20} {}'.format(gof, value)
+                         for gof, value in sorted(self.gof_qualifiers.items()))
```
pckroon: You can make the `:20` dependent on the largest length of any GoF qualifier ;) See the comment above.

symfit/core/fit_results.py
```diff
@@ -155,4 +176,43 @@ def _array_safe_dict_eq(one_dict, other_dict):
...
+        - Likelihood and log-likelihood for
+          :class:`~symfit.core.objectives.LogLikelihood`.
+
+        :return: ``dict`` containing goodness of fit qualifiers.
+        """
+        gof_qualifiers = {}
+
+        gof_qualifiers['objective_value'] = self.minimizer_output.fun
+        if isinstance(self.objective, (LeastSquares, VectorLeastSquares)):
+            from symfit.core.fit import r_squared
```
pckroon: Move r_squared to this file instead

symfit/core/fit_results.py
```diff
@@ -155,4 +176,43 @@ def _array_safe_dict_eq(one_dict, other_dict):
...
+            # In this case the objective value is the residuals
+            chi_squared = np.sum(gof_qualifiers['objective_value'] ** 2)
+            gof_qualifiers['chi_squared'] = chi_squared
+        elif isinstance(self.objective, LeastSquares):
+            # Undo the normalization to get back chi^2.
+            gof_qualifiers['chi_squared'] = 2 * gof_qualifiers['objective_value']
+        elif isinstance(self.objective, LogLikelihood):
+            # We undo the minus sign we have included to maximize likelihood
+            gof_qualifiers['log_likelihood'] = - gof_qualifiers['objective_value']
+            gof_qualifiers['likelihood'] = np.exp(- gof_qualifiers['objective_value'])
```
pckroon: `np.exp(gof_qualifiers['log_likelihood'])` is slightly clearer in intent.

symfit/core/minimizers.py
```diff
@@ -270,7 +271,7 @@ def execute(self, **minimizer_kwargs):
         # TODO: Compile all previous results in one, instead of just the
         # number of function evaluations. But there's some code down the
         # line that expects scalars.
-        final.infodict['nfev'] = sum(ans.infodict['nfev'] for ans in answers)
```
pckroon: It would be nice if Objectives kept track of the number of (function|jacobian|hessian) evaluations. Different PR though.

symfit/core/fit_results.py
tBuLi: Good point, that is definitely not the intent. Internally we do assume this object for now, and in the near feature we pick attributes from `minimizer_output` based on the type of `minimizer`.

pckroon: How about feeding minimizer output as keywords to FitResult? That way the unpacking from minimizer output becomes the responsibility of Minimizer objects, and it will still be compatible with frameworks that produce dicts.

tBuLi: You mean that `minimizer_output` is a dict, or that all minimizer output should be fed to FitResults as keywords? I noticed that the latter was becoming a mess, which is why I switched to this, but making minimizer_output into a standardized dict might be nice. This is also in fact what they finally have decided to do at scipy, because the more recent version of `leastsq` now also returns a `OptimizeResult`, which is just a dict with fancy printing. However, we don't get that yet because we use the `leastsqbound` package here. Switching to scipy's current support for bounds is another future PR.

pckroon: > You mean that minimizer_output is a dict, or that all minimizer output should be fed to FitResults as keywords? I noticed that the latter was becoming a mess, which is why I switched to this, but making minimizer_output into a standardized dict might be nice.

I mean the latter. It might make for a very long signature, but it makes documenting what's expected, as well as default arguments easier. The options are functionally equivalent though.

tBuLi: The problem I see with that is that even now not every minimizer returns the same stuff, so then we would have to add a lot of arguments to the signature which are usually not provided. Plus I think it's good to have access to the raw output just in case, because we might not always be up to date with changes made to `scipy` or other libraries in the future. Therefore, I think just accepting a dict with all the output is a good way to get around that.

pckroon: Why not both? :)
We can give all arguments used by FitResult a name, and accept everything else as `**kwargs`. That way you can always construct a FitResult with your **minimizer_output, and the API of FitResult becomes fairly clean in what should be provided at minimum. An helper method to rename (some) dict keys might be valuable, but that's for later.

tBuLi: Good point, I agree that we should keep mandetory arguments separate, and now that `gof_qualifiers` are calculated by FitResults itself, the kwargs could be the remaining minimizer output.

symfit/core/fit_results.py
```diff
@@ -46,15 +47,20 @@ def __init__(self, model, popt, covariance_matrix, minimizer_output, minimizer,
 
     @property
     def iterations(self):
-        return self.minimizer_output.get('nit', None)
+        if 'nit' in self.minimizer_output:
+            return self.minimizer_output['nit']
+        elif 'niter' in self.minimizer_output:
+            return self.minimizer_output['niter']
```
pckroon: I really like this being a property, but I'd prefer to leave the translation to Minimizers. It's not super important though

symfit/core/fit_results.py
```diff
@@ -171,13 +172,18 @@ def _array_safe_dict_eq(one_dict, other_dict):
...
                 elif key == 'minimizer_output':
                     # Ignore this, because it can contain unexpected terms and
                     # if all the derived attributes are correct I see no reason
                     # why this term shouldn't be at least close enough.
                     pass
+                elif key == 'constraints':
+                    assert len(one_dict['constraints']) == len(other_dict['constraints'])
+                    for constr1, constr2 in zip(one_dict['constraints'],
+                                                other_dict['constraints']):
+                        assert constr1 == constr2
```
pckroon: How is this different from `assert one_dict['constraints'] == other_dict['constraints']`? And should the order of the constraints matter?

symfit/core/objectives.py
```diff
@@ -124,6 +124,26 @@ def _invariant_kwargs(self):
         )
         return kwargs
 
+    def __eq__(self, other):
+        """
+        Objectives are considered equal if they are of the same type, have the
+        same model, and the same data.
+        """
+        if self.__class__ == other.__class__ and self.model == other.model:
```
pckroon: Why not allow subclassing?

symfit/core/fit_results.py
tBuLi: I agree that it would be better for all the minimizers to return something consistent in the future, but indeed this does not affect the API at the moment so I'm fine with this.

(p.s. for trust-constr it is suddenly called 'niter'... It's almost seems like a sport at scipy to be inconsistent.)

symfit/core/fit_results.py
tBuLi: Hmmm, it no longer is. That's because I first hadn't defined `__eq__` for objectives. Good catch.

symfit/core/objectives.py
tBuLi: Not sure, should a subclass of an objective be the same as the original so long as it has the same data and model? I don't think so, right?

pckroon: That's the question. It's something to think about. Maybe just drop the class equality, and only compare attributes?

symfit/core/fit_results.py
tBuLi: I'll update it to a set compare to ignore ordering as well.

symfit/core/objectives.py
tBuLi: I think the false positives one could get then are way worse. For example, even if I make a small modification like a least squares which returns the absolute magnitude in case of complex numbers, I would already not want that to compare as the same. So I think the class equality makes sense.

pckroon: Ok, add a comment with this reasoning, and leave it :)

symfit/core/fit_results.py
```diff
@@ -10,36 +15,52 @@ class FitResults(object):
...
+    - objective and minimizer used.
 
     Contains the attribute `params`, which is an
     :class:`~collections.OrderedDict` containing all the parameter names and
     their optimized values. Can be `**` unpacked when evaluating
     :class:`~symfit.core.fit.Model`'s.
     """
-    def __init__(self, model, popt, covariance_matrix, infodic, mesg, ier, **gof_qualifiers):
+    @keywordonly(constraints=None)
+    def __init__(self, model, popt, covariance_matrix, minimizer, objective, **minimizer_output):
```
pckroon: My current problem with this signature is that `minimizer_ouput` *must* contain a `'message'` and `'infodic'`, which is rather obscure to figure out. I suggest to explicitly add those to to the signature.

symfit/core/fit_results.py
pckroon: I think it would be a fairly small change

symfit/core/fit_results.py
```diff
@@ -147,4 +191,73 @@ def _array_safe_dict_eq(one_dict, other_dict):
...
     def __eq__(self, other):
-        return FitResults._array_safe_dict_eq(self.__dict__, other.__dict__)
\ No newline at end of file
+        return FitResults._array_safe_dict_eq(self.__dict__, other.__dict__)
+
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        state['minimizer'] = (type(state['minimizer']),
+                              state['minimizer'].objective,
+                              state['minimizer'].parameters)
```
pckroon: Does this include parameter initial values?

symfit/core/fit_results.py
```diff
@@ -147,4 +191,73 @@ def _array_safe_dict_eq(one_dict, other_dict):
...
+        (g.o.f.) qualifiers.
+
+        The ``objective_value`` itself always exists, and then depending on the
+        objective we also get the following:
+
+        - The coefficient of determination :math:`R^2` and :math:`\\chi^2` for
+          :class:`~symfit.core.objectives.LeastSquares` and
+          :class:`~symfit.core.objectives.VectorLeastSquares`.
+        - Likelihood and log-likelihood for
+          :class:`~symfit.core.objectives.LogLikelihood`.
```
pckroon: Can't you always calculate all? :)

symfit/core/minimizers.py
```diff
@@ -511,6 +501,11 @@ def scipy_constraints(self, constraints):
         cons = tuple(cons)
         return cons
 
+    def _pack_output(self, ans):
+        fit_result = super(ScipyConstrainedMinimize, self)._pack_output(ans)
+        fit_result.constraints = self.constraints
```
pckroon: No. Bad. Shouldn't those be part of the Model anyway? Let FitResult take it from there.

symfit/core/fit_results.py
tBuLi: No I think that's a bit dangerous since they don't all have meaning at the same time. So I'd rather err on the side of caution, then risk having someone cite a number that doesn't actually exist for their problem. If they disagree for their particular problem they can always go ahead and calculate it manually :).

symfit/core/minimizers.py
tBuLi: constraints are not a part of the model, but of `Fit` and/or the minimizer. So I don't see why this is particularly bad? Python is a consenting adults language? :)

symfit/core/fit_results.py
tBuLi: That question is work in progress ;). Still, the equality checking in the pickle testing passes successfully, so I'm going to say yes?

symfit/core/fit_results.py
pckroon: Works for me. Future PR would be to expose those functions somewhere.

symfit/core/minimizers.py
pckroon: Shouldn't constraints be part of the Model? (I feel like we've had that discussion before).
Either way, I also don't immediately see a better way of solving htis.

symfit/core/fit_results.py
tBuLi: Good point, I've added `message` to the signature. However, `infodic` is something SciPy specific, so I have actually removed that as a `property`. It is still accessible through `minimizer_output` but I think that is good enough.

tBuLi:
Chained minimizers are not correctly shown as the minimizer, but instead the last is. Should be the list used instead.

symfit/core/minimizers.py
tBuLi: No they aren't, but any BaseModel can be used as a constraint now.

tests/test_minimizers.py
```diff
@@ -27,6 +27,35 @@ def chi_squared(x, y, a, b, sum=True):
...
+        chi2_jac = super(SqrtLeastSquares, self).eval_jacobian(*args, **kwargs)
+        return 0.5 * (1 / sqrt_chi2) * chi2_jac
+
+    def eval_hessian(self, *args, **kwargs):
+        sqrt_chi2 = self(*args, **kwargs)
+        sqrt_chi2_jac = self.eval_jacobian(*args, **kwargs)
+        chi2 = super(SqrtLeastSquares, self).__call__(*args, **kwargs)
+        chi2_jac = super(SqrtLeastSquares, self).eval_jacobian(*args, **kwargs)
+        chi2_hess = super(SqrtLeastSquares, self).eval_hessian(*args, **kwargs)
+        # Hessian Approximation! Just takes the
```
pckroon: missing some words?

tests/test_minimizers.py
```diff
@@ -27,6 +27,35 @@ def chi_squared(x, y, a, b, sum=True):
...
+        raise - 0.5 * (1 / chi2) * np.outer(sqrt_chi2_jac, chi2_jac) \
```
pckroon: should be return, not raise?

symfit/core/argument.py
```diff
@@ -115,6 +115,23 @@ def __init__(self, name=None, value=1.0, min=None, max=None, fixed=False, **assu
...
 
+    def __eq__(self, other):
+        """
+        Parameters are considered equal when their name, assumptions, and
+        bounds are considered the same.
+        """
+        equal = super(Parameter, self).__eq__(other)
+        if equal is NotImplemented:
+            return equal
+        else:
```
pckroon: Pointless else

symfit/core/fit.py
```diff
@@ -527,9 +527,14 @@ def connectivity_mapping(self, value):
...
 
     def __eq__(self, other):
-        raise NotImplementedError(
-            'Equality checking for {} is ambiguous.'.format(self.__class__.__name__)
-        )
+        if self.connectivity_mapping == other.connectivity_mapping:
+            for key, func in self.model_dict.items():
+                if func != other[key]:
+                    return False
+            else:
```
pckroon: Pointless else

symfit/core/fit.py
```diff
@@ -527,9 +527,14 @@ def connectivity_mapping(self, value):
...
+                return True
+        else:
```
pckroon: Pointless else

symfit/core/fit.py
```diff
@@ -527,9 +527,14 @@ def connectivity_mapping(self, value):
         self._connectivity_mapping = value
 
     def __eq__(self, other):
-        raise NotImplementedError(
-            'Equality checking for {} is ambiguous.'.format(self.__class__.__name__)
-        )
+        if self.connectivity_mapping == other.connectivity_mapping:
```
pckroon: I would invert this condition.
```python3
if self.con_map != other.con_map:
    return False
for k, v in self.model_dict.items():
    ...
return True
```

symfit/core/fit_results.py
```diff
@@ -146,5 +187,86 @@ def _array_safe_dict_eq(one_dict, other_dict):
...
+        min_class, objective, parameters = state['minimizer']
+        try:
+            # If min_class is iterable, initiate a ChainedMinimizer.
+            minimizers = [cls(objective, parameters) for cls in min_class]
+        except TypeError:
+            state['minimizer'] = min_class(objective, parameters)
+        else:
+            # Has to be imported here to prevent circular imports
+            from symfit.core.minimizers import ChainedMinimizer
+            state['minimizer'] = ChainedMinimizer(objective, parameters,
```
pckroon: use `min_class` instead?

symfit/core/fit_results.py
```diff
@@ -146,5 +187,86 @@ def _array_safe_dict_eq(one_dict, other_dict):
...
+        state['minimizer'] = (minimizer_cls,
+                              state['minimizer'].objective,
+                              state['minimizer'].parameters)
+        return state
+
+    def __setstate__(self, state):
+        min_class, objective, parameters = state['minimizer']
+        try:
+            # If min_class is iterable, initiate a ChainedMinimizer.
+            minimizers = [cls(objective, parameters) for cls in min_class]
```
pckroon: Shouldn't this be `min_class.minimizers`?

symfit/core/objectives.py
```diff
@@ -124,6 +124,30 @@ def _invariant_kwargs(self):
...
+                try:
+                    equal = np.allclose(other.data[key], value)
+                except TypeError:
+                    equal = other.data[key] == value
+                finally:
+                    if not equal:
+                        return False
+            else:
+                return True
+        else:
```
pckroon: Useless `else` statements. Also, invert the condition above: `if not equal classes: return False`

symfit/core/fit_results.py
tBuLi: `min_class` is an iterable of minimizer types in this case, corresponding to the `.minimizers` of the original ChainedMinimizer.

tests/test_minimizers.py
tBuLi: I removed it all together, it was outdated since I actually wrote down the full hessian.

symfit/core/fit_results.py
tBuLi: turns out that ChainedMinimizer already works directly, so this can be simplified. Stay tuned.

symfit/core/argument.py
pckroon: I would still invert this condition: 
```python3
if not equal:
    return False
return (self.min == other.min and
        self.max == other.max and
        self.fixed == other.fixed and
        self.value == other.value)
```

symfit/core/objectives.py
tBuLi: I'm not sure I follow on the inversion. However, I will remove the last else and do a break of the loop instead, so the for/else really serves a pourpose and `False` is our catch-all.

symfit/core/minimizers.py
tBuLi: I agree that would be nice, but indeed it should be a different PR. The important thing is that now at least we catch all results under `minimizer_output`, so at least it's there if you really need it.

tBuLi:
I'm now very happy with this PR as it stands. There is always more work to be done, but this PR makes sure all the information I want during fitting is in one easy to print report.

symfit/core/argument.py
pckroon: Still a pointless else :D

symfit/core/objectives.py
```diff
@@ -142,11 +142,10 @@ def __eq__(self, other):
                     equal = other.data[key] == value
                 finally:
                     if not equal:
-                        return False
+                        break
             else:
                 return True
-        else:
-            return False
+        return False
```
pckroon: My suggestion:
```python3
if self.__class__ != other.__class__ or self.model != other.model:
    return False

for key, value in self.data.items():
    try:
        equal = np.allclose(other.data[key], value)
    except TypeError:
        equal = other.data[key] == value
    finally:
        if not equal:
            return False
return True

```

pckroon:
Two more nitpicks. After that I having nothing left to complain about 👍
Nice work!

PS. well, maybe the drop in coverage. But I'll leave that up to you. Might be better in a separate PR

symfit/core/argument.py
tBuLi: I'm personally more of a fan of having the else there (as you probably guessed :P) because I think it improves readability.

symfit/core/objectives.py
tBuLi: I think it's harder to read with all the negations. I personally prefer to have if conditionals without `not` type operations.

pckroon: The difference is that I don't need the for/else.
In my example you very obviously try to find the short-circuit/easy answer: when they're not equal.

symfit/core/argument.py
pckroon: In this case it's not too important. But keeping the indentation under control starts to matter for larger blocks of code and nested constructs

symfit/core/objectives.py
tBuLi: Fair enough, and there something to be set for keeping indentation under control

PR #224, opened 2019-04-02
Create first hypothesis test

pckroon:
This is the first step into pulling the tests into 2019 by 1) introducing pytest, and more importantly 2) setting up Hypothesis style testing.
The first test generates 500 random models with associated data, and tries to fit with them.

TODO:
- [x] Generate Models
- [ ] Constraints
- [ ] Parameter bounds
- [ ] ODE Models
- [ ] Pick random minimizers
- [ ] Pick random objectives
- [ ] Matrix variables/parameters/models/whatevers? I don't know well enough how this (should) work to do this.
- [ ] More rigorous answer validation

### DIFF
--- a/symfit/core/fit.py
+++ b/symfit/core/fit.py
@@ -286,8 +286,14 @@ class BaseModel(Mapping):
         # Everything at the bottom of the toposort is independent, at the top
         # dependent, and the rest interdependent.
         ordered = list(toposort(self.connectivity_mapping))
-        independent = sorted(ordered.pop(0), key=sort_func)
-        self.dependent_vars = sorted(ordered.pop(-1), key=sort_func)
+        if ordered:
+            independent = sorted(ordered.pop(0), key=sort_func)
+        else:
+            independent = []
+        if ordered:
+            self.dependent_vars = sorted(ordered.pop(-1), key=sort_func)
+        else:
+            self.dependent_vars = []
         self.interdependent_vars = sorted(
             [item for items in ordered for item in items],
             key=sort_func
--- a/tests/hypothesis_helpers.py
+++ b/tests/hypothesis_helpers.py
@@ -1,0 +1,295 @@
+from functools import reduce
+import operator
+
+import numpy as np
+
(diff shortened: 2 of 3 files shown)


pckroon:
By the way, see also: https://hypothesis.readthedocs.io/en/latest/quickstart.html on what and why we want it.

pckroon:
So, some more got done. This makes a semi-random model, and tries fitting it with the answer as initial guess. This should make sure you can call Fit with pretty much any model and data without it blowing up in your face. In addition, it make sure the minimizers don't walk away from the correct answer.

Second, this makes polynomial models with the parameters as coefficients. These are actually fitted with arbitrary initial guesses (even out of bounds!), letting Fit figure out whether it should use BFGS or LBFGSB. That method does have serious issues with numerical stability though, and there are a lot of assumption/filters around to make sure the data is not too small or large. Otherwise you run the risk of the minimizer happily terminating with either a gradient convergence method, or a too small relative step in the objective.

Left to do are fixed parameters (should be pretty easy), constraints (really hard, since we need to know a/the feasible answer), ODEModels (somewhat hard), Objectives (easy-ish), and Minimizers (somewhat easy). It might also be a good idea to use this to validate the `finite_difference` method. But that will also have *serious* numerical issues, so let's leave that for later.

@tBuLi In how many ways can you provide constraints? Are there any restriction in what parameters/variables I can use to make those?

Lastly, this PR still needs *serious* cleaning before it's ready to be merged.

tBuLi:
Thanks for this PR, looks challenging but cool.

To answer your question, nowadays any model can be used as a constraint: they can even contain variables which are not in the original model! The only restriction is that their parameters are a subset of those of the model. Then, the constraint object will be made to have all the parameters of the model so it can be called consistently.

pckroon:
But those models don't need to be instantiated yet right? I could also use simple (relational) expressions. What I need it a concise/conclusive list of all options.

pckroon:
This is too old, and needs some work to be resurrected. Will be back :)

PR #225, opened 2019-04-03
Update Python versions for Travis. Remove Sphinx version pinning

pckroon:
Py3.4 is no longer supported and removed from Travis. In addition, Py3.7 is added.

Also, it seems that pinning sphinx to 1.5.5 is no longer necessary. Locally I can run sphinx 2.0.0 in nitpicky mode without issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+dist: xenial
 python:
   - "2.7"
   - "3.5"
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,9 @@ python:
   - "3.7"
 cache:
   - pip
-before_install:
-## Provide a X display for plotting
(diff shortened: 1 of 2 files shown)


pckroon:
Also, for your consideration: I can also set up build-stages, for example: https://travis-ci.org/marrink-lab/vermouth-martinize

tBuLi:
I've had a quick look and the stages do make everything look a lot more presentable so if you would be willing to do that, you have my full support!

Also, how the hell does this decrease coverage? The wonders of coverage.

pckroon:
As to the decrease in coverage I have a suspicion:
In construct like this:
```
for ...:
   if ...:
       continue/break
```
or `for: else: continue` python optimizes away the execution of the `continue/break` line. So coverage never sees them as being executed. My guess is that this optimization became more powerful in 3.7.

Issue #215, opened 2019-02-11
forced ODE


vrutten:
Hi,

Thank you for your great framework.
I am trying to fit a forced ODE (where the forcing occurs at a known time).
e.g.: data in the plot below, where a step input comes at known times.
<img width="631" alt="screenshot 2019-02-10 at 23 18 02" src="https://user-images.githubusercontent.com/15912669/52546168-0d65f300-2d8b-11e9-866c-8f4120771c48.png">

How would you recommend I go about parameterising this?
I tried a few different ways but non would run.

```
y, h, t = variables('y, h, t')
tau, b = parameters('tau, b')
model_dict = {D(y, t): (- y + b + h)/tau} 
```
and then have a CallableNumericalModel?
or
```
y, t = variables('y, t')
tau, b = parameters('tau, b')
model_dict = {D(y, t): (- y + b + h(t))/tau} 
```
where h itself is a function
e.g.:
```
def h(t):
    if t>2 and t<5:
        return 1
    else:
         return 0
```

Thanks for any help - it would really be appreciated.

best,

V

pckroon:
Hi V,

that would probably work :) (read: should)

You'll probably get better results however, if you can come up with continuous form for `h`, for example the product of two steep sigmoids.

```python3
def sigmoid(t, t0=0, k=1, N=1):
    return N/(1+symfit.exp(-k*(t-t0)))

def h(t):
    return sigmoid(t, 2, 100, 1) * (1 - sigmoid(t, 5, 100, 1))
```

Experiment and adjust as needed ;)

tBuLi:
Don't forget that you need to use `ODEModel` in order for the system to be integrated, otherwise indeed it won't run. And I think you could replace `h(x)` by either a `sympy.Piecewise`, or a combination of `sympy.Heaviside`.

pckroon:
> Don't forget that you need to use `ODEModel` in order for the system to be integrated...

Doesn't `Fit` automagically determine there's a derivative in `model_dict`, and just Do The Right Thing (tm)? If not, that's a bug :)

> And I think you could replace `h(x)` by either a `sympy.Piecewise`, or a combination of `sympy.Heaviside`.

I'm still utterly convinced taking a function that is actually derivable makes everything smoother (pun intended). It could very well be that the ODE integrator doesn't like the sudden change either way. IIRC the scipy integrators are fairly intelligent in dealing with steep region by automatically reducing the step size. But that doesn't help for step functions.

[...]

PR #229, opened 2019-04-14
Fix 215

tBuLi:
In response to #215, I added a simple warning to prevent this confusion in the future.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,5 +4,4 @@ sympy >= 1.2
 toposort
 funcsigs; python_version < '3.0'
 functools32; python_version < '3.0'
-unittest2; python_version < '3.2'
 six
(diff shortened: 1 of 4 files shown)


tBuLi:
That's a much bigger job than this PR. I wouldn't call unittest2 a third party app, it's just a backport of newer unittest facilities that don't exist in py27 but make supporting it much easier. I don't think it is worth it to deal with py27 warning handling for testing for the coupe of py27 users left (hopefully).

pckroon:
I don't intend this PR to redo all the tests. Instead, just use e.g. `with pytest.warns` rather than unittest2:
https://docs.pytest.org/en/latest/warnings.html

tBuLi:
Ah, that changes things. That's a good idea, I'll change it.

PR #230, opened 2019-04-14
Fix 214

### LINKED ISSUE #214: Better sanity checking on input data
Consider the following test cases:

```python
A, mu, sig = parameters('A, mu, sig')
x, y, Y = variables('x, y, Y')

model = Model({y: A * Gaussian(x, mu=mu, sig=sig)})

np.random.seed(2)
xdata = np.random.normal(1.2, 2, 10)
ydata, xedges = np.histogram(xdata, bins=int(np.sqrt(len(xdata))),
                             density=True)

# Allowed
fit = Fit(model, x=xedges[:-1], y=ydata)
fit = Fit(model, x=xdata, objective=LogLikelihood)

# Not allowed
with self.assertRaises(TypeError):
    fit = Fit(model, x=xedges[:-1], y=ydata, Y=2)
with self.assertRaises(TypeError):
    fit = Fit(model, x=xdata)
with self.assertRaises(TypeError):
    fit = Fit(model, x=xdata, y=ydata, objective=LogLikelihood)
```
The last two of these currently don't give the expected error, because all variables that have not been provided are set to `None` by default. This has to be changed, because it can make debugging very hard if you forget to provide a variable with data. However, this design choise was made so we can write things like:
```python
x = Parameter('x')
y = Variable('y')

model = Model({y: x**4 - 10 * x**2 - x})
fit = Fit(model)
```
instead of having to write 
```python
fit = Fit(model, y=None)
```

We should either include some logic to still accept the former case, or we make an API change where users will have to write the latter.

tBuLi:
Fixed #214. Variables are no longer set to None by default, all variables now need to be provided explicitly with data or None, unless `Fit` can infer from the data, model and objective provided that some should be ignored. This way debugging will be greatly improved, since currently it is still allowed to forget to provide data which causes a lot of pain when debugging.

Only important API change:  `None` has to be provided explicitly to `Fit` when no data is present for a variable.

### DIFF
--- a/symfit/core/fit.py
+++ b/symfit/core/fit.py
@@ -989,16 +989,6 @@ class TakesData(object):
             else:
                 self.absolute_sigma = False
 
-    def _make_parameters(self):
-        """
-        Make :class:`inspect.Parameter`'s using the default rules: all variables
-        to the model need to be provided.
-
-        :return: list of  :class:`inspect.Parameter` corresponding to
-            ``self.model``.
-        """
-        return self._make_signature_params_for_objective(self.model, BaseObjective)
-
     def _make_signature(self):
         """
         Make a :class:`inspect.Signature` object corresponding to
@@ -1007,31 +997,33 @@ class TakesData(object):
         :return: :class:`inspect.Signature` object corresponding to
             ``self.model``.
         """
-        parameters = self._make_parameters()
+        parameters = self._make_parameters(self.model)
         parameters = sorted(parameters, key=lambda p: p.default is None)
         return inspect_sig.Signature(parameters=parameters)
 
     @staticmethod
-    def _make_signature_params_for_objective(model, objective_cls):
+    def _make_parameters(model, none_allowed=None):
         """
-        Based on an objective function, return the inspect.Parameter objects
-        needed to satisfy this objective.
+        Based on a model, return the inspect.Parameter objects
+        needed to satisfy all the variables of this model.
 
         :param model: instance of model
-        :param objective_cls: (subclass) of objective.
+        :param none_allowed: If provided, this has to be a sequence of
+            :class:`symfit.core.argument.Variable` whose values are set to
+            ``None`` by default. If not provided, this will be set to sigma
+            variables only.
+        :return: list of  :class:`inspect.Parameter` corresponding to all the
+            external variables of the model.
         """
-        if issubclass(objective_cls, BaseIndependentObjective):
-            vars = model.independent_vars
-        else:
-            vars = model.vars
-
+        if none_allowed is None:
+            none_allowed = model.sigmas.values()
         parameters = [
             inspect_sig.Parameter(
                 var.name,
                 kind=inspect_sig.Parameter.POSITIONAL_OR_KEYWORD,
-                default=None if var in model.sigmas.values() else inspect_sig.Parameter.empty
+                default=None if var in none_allowed else inspect_sig.Parameter.empty
             )
-            for var in vars
+            for var in model.vars
         ]
         return parameters
 
@@ -1243,7 +1235,8 @@ class Fit(HasCovarianceMatrix):
         fit_result = fit.execute(minimizer_kwargs=[dict(xatol=0.1), {}])
     """
 
-    @keywordonly(objective=None, minimizer=None, constraints=None)
+    @keywordonly(objective=None, minimizer=None, constraints=None,
+                 absolute_sigma=None)
     def __init__(self, model, *ordered_data, **named_data):
         """
 
@@ -1271,6 +1264,7 @@ class Fit(HasCovarianceMatrix):
         objective = named_data.pop('objective')
         minimizer = named_data.pop('minimizer')
         constraints = named_data.pop('constraints')
+        absolute_sigma = named_data.pop('absolute_sigma')
         # Should be a list of Constraint objects
         constraints = [] if constraints is None else constraints
 
@@ -1280,38 +1274,29 @@ class Fit(HasCovarianceMatrix):
         else:
             self.model = Model(model)
 
-        # Select objective function to use. Has to be done before calling
-        # super.__init__
-        self.objective = self._determine_objective(self.model, objective,
-                                                   minimizer)
         self.constraints = self._init_constraints(constraints=constraints,
                                                   model=self.model)
-        try:
-            super(Fit, self).__init__(model, *ordered_data, **named_data)
-        except TypeError as err:
-            # There is one possible exception: if the objective was determined
-            # to be a MinimizeModel, but the user actually provided data for the
-            # dependent vars. In that case, they actually want a LeastSquare so
-            # we retry.
-            if objective != self.objective and 'argument' in str(err):
-                self.objective = LeastSquares
-                try:
-                    super(Fit, self).__init__(model, *ordered_data, **named_data)
-                except TypeError:
-                    # If this did not turn out to be the case, re-raise the
-                    # original exception
-                    raise err
-            else:
-                # However, if the user provided an objective themselves,
-                # or it occurred for another reason, they'll want this error.
-                raise err
+
+        # Bind as much as possible the provided arguments.
+        signature = self._make_signature()
+        bound_arguments = signature.bind_partial(*ordered_data, **named_data)
+
+        # Select objective function to use. Has to be done before calling
+        # super.__init__
+        self.objective = self._determine_objective(
+            self.model, objective=objective,
+            minimizer=minimizer, bound_arguments=bound_arguments
+        )
+
+        super(Fit, self).__init__(self.model, absolute_sigma=absolute_sigma,
+                                  **bound_arguments.arguments)
 
         # Update the data belonging to the constraints. We do this by checking
         # for the presence of data with the same name as one of the independent
         # variables of the constraint. If present, we start addressing them by
         # their Variable instead.
         for constraint in self.constraints:
-            for var in constraint.independent_vars:
+            for var in constraint.vars:
                 if var.name in self.data:
                     self.data[var] = self.data.pop(var.name)
 
@@ -1331,20 +1316,16 @@ class Fit(HasCovarianceMatrix):
             self.minimizer = self._init_minimizer(minimizer)
 
     def _make_signature(self):
-        if issubclass(self.objective, BaseObjective):
-            objective_cls = self.objective
-        else:
-            objective_cls = self.objective.__class__
-        parameters = self._make_signature_params_for_objective(
-            self.model, objective_cls
-        )
+        parameters = self._make_parameters(self.model)
         # Extend the signature with the variables to the constraint. Since
         # constraints will be turned into MinimizeModel objectives, they only
         # need independent variables to be provided.
         for constraint in self.constraints:
+            none_allowed = constraint.dependent_vars + list(constraint.sigmas.values())
             parameters.extend(
-                self._make_signature_params_for_objective(constraint,
-                                                          MinimizeModel)
+                self._make_parameters(
+                    constraint, none_allowed=none_allowed
+                )
             )
 
         # Make unique while preserving order, and sort by default value so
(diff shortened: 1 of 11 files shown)


symfit/core/fit.py
```diff
@@ -1244,13 +1271,40 @@ def __init__(self, model, *ordered_data, **named_data):
...
-                                                  model=model)
-        super(Fit, self).__init__(model, *ordered_data, **named_data)
+                                                  model=self.model)
+        try:
+            super(Fit, self).__init__(model, *ordered_data, **named_data)
+        except TypeError as err:
+            # There is one possible exception: if the objective was determined
+            # to be a MinimizeModel, but the user actually provided data for the
+            # dependent vars. In that case, they actually want a LeastSquare so
+            # we retry.
```
pckroon: Doesn't this mean that objective auto-detection should be smarter, rather than trying to fix it up afterwards?

symfit/core/fit.py
```diff
@@ -1448,10 +1509,11 @@ def r_squared(model, fit_result, data):
...
     """
     # First filter out the dependent vars
-    y_is = [data[var] for var in model if var in data]
-    x_is = [value for var, value in data.items() if var.name in model.__signature__.parameters]
+    y_is = [data[var] for var in model.dependent_vars if var in data]
+    x_is = [data[var] for var in model.independent_vars if var in data]
     y_bars = [np.mean(y_i) if y_i is not None else None for y_i in y_is]
-    f_is = model(*x_is, **fit_result.params)
+    f_is = model(*x_is, **fit_result.params)._asdict()
+    f_is = [f_is[var] for var in model.dependent_vars]
```
pckroon: I have the idea this line is a no-op

symfit/core/objectives.py
```diff
@@ -124,6 +127,17 @@ def _invariant_kwargs(self):
         )
         return kwargs
 
+    def _sanity_checking(self):
+        """
+        Check if the model and the provided data are compatible. Raises a
+        TypeError when this is not the case.
+        """
+        # Simply checking for existence of these dicts will raise an error if
+        # they cannot be built.
```
pckroon: extra comment: because they are properties.

symfit/core/fit.py
tBuLi: Yeah I agree in principle, but I have been unable to do that so far. The main problem is that the determination depends on both the objective and the data. `TakesData` does not know about objectives, it only matches the data to the model. I feel like it isn't right to pass the objective to the `TakesData` super, but if we don't then the signature will not be satisfied and thus raise an error.

One way around this might be to have a seperate class/function which checks the signature based on the model, data and optionally the objective, which both TakesData and FIt can call in their own way.

symfit/core/fit.py
tBuLi: Not if the model has `interdependent_vars`. Since ans really evaluates all components of the model, it will also contain the ones corresponding to the interdependent variables. API wise I think that is fine and consistent with the model basicly being its model_dict, but it means we should ignore them here.

pckroon: aaah.... I missed that edge case indeed :)
Maybe: `f_is = [f_is[var] for var in f_is if var not in model.interdependent_vars]`?

symfit/core/fit.py
pckroon: I'm slightly confused on who calls what now, but that's ok.
My initial goal would be to keep all the "intelligence" in Fit, which knows data and model, and can therefor select the appropriate Objective. Who is missing information then?

symfit/core/fit.py
tBuLi: I agree that it would be more readable, but I do feel like since there are more steps in between, I can more easily see the ordering go wrong with future bugs by the time we get to the `Ans` object. But I'll add a comment, that should also do the trick.

symfit/core/fit.py
```diff
@@ -1243,7 +1235,8 @@ class Fit(HasCovarianceMatrix):
         fit_result = fit.execute(minimizer_kwargs=[dict(xatol=0.1), {}])
     """
 
-    @keywordonly(objective=None, minimizer=None, constraints=None)
+    @keywordonly(objective=None, minimizer=None, constraints=None,
+                 absolute_sigma=None)
```
pckroon: The default value for `absolut_sigma` should be True, rather than None

symfit/core/fit.py
```diff
@@ -1372,31 +1353,52 @@ def _determine_minimizer(self):
...
+                    isinstance(objective, (MinimizeModel, LogLikelihood)):
+                # Set dependent vars and corresponding sigmas to None.
+                for var in model.dependent_vars + list(model.sigmas.values()):
+                    if var.name not in bound_arguments.arguments:
+                        bound_arguments.arguments[var.name] = None
+                    else:
+                        raise TypeError(
+                            'A value was provided for `{}`, however for {} '
+                            'fitting the dependent variable cannot have a value '
+                            'assigned to it.'.format(objective, var.name)
```
pckroon: Formatting is the wrong way round. Should be var.name first, objective second.

symfit/core/fit.py
```diff
@@ -1372,31 +1353,52 @@ def _determine_minimizer(self):
...
+                # be considered separately and has its own non standard
                 # objective function.
                 return VectorLeastSquares
-            # Param only scalar Model -> the model is the objective.
-            elif len(model.independent_vars) == 0 and len(model) == 1:
-                # No data provided means a simple minimization of the Model
-                # parameters is requested, not a fit.
-                return MinimizeModel
+            if len(model) == 1:
+                if len(model.independent_vars) == 0 and model.dependent_vars[0].name not in bound_arguments.arguments:
```
pckroon: Combine the conditions?

symfit/core/fit.py
tBuLi: Good suggestion as I first thought you were right, but it turns out that in `TakesData` it is actually also None, and then determined there. (Usually true by default indeed.)

symfit/core/fit.py
tBuLi: I decided to go for two steps otherwise this line would become way to long, even for my taste. But I will reordered them slightly to make it even more compact.

symfit/core/fit.py
```diff
@@ -1390,8 +1390,8 @@ def _determine_objective(model, objective, minimizer, bound_arguments):
                 # be considered separately and has its own non standard
                 # objective function.
                 return VectorLeastSquares
-            if len(model) == 1:
-                if len(model.independent_vars) == 0 and model.dependent_vars[0].name not in bound_arguments.arguments:
+            if len(model) == 1 and len(model.independent_vars) == 0:
+                if model.dependent_vars[0].name not in bound_arguments.arguments:
```
pckroon: I still think it's better to combine them:
```
if (len(model) == 1 and  not model.independent_vars
        and model.dependent_vars[0].name not in bound_arguments.argument):
```
Or make a variable out of some of them.

tBuLi: You're right, that is better. After reading PEP 8 on this it seems like that is the solution. PyCharm automatically breaks long lines with `\`, which I really dislike and try to avoid at all cost. But I forgot the parentheses solve this problem in a very clean way.

symfit/core/fit.py
tBuLi: This is solved now.

symfit/core/fit.py
```diff
@@ -1390,15 +1390,15 @@ def _determine_objective(model, objective, minimizer, bound_arguments):
...
-                if model.dependent_vars[0].name not in bound_arguments.arguments:
-                    # No data provided means a simple minimization of the Model
-                    # parameters is requested, not a fit. Therefore set the
-                    # variable to None and return MinimizeModel.
-                    for var in model.dependent_vars:
-                        if var.name not in bound_arguments.arguments:
-                            bound_arguments.arguments[var.name] = None
-                    return MinimizeModel
+            if (len(model) == 1 and len(model.independent_vars) == 0 and
+                model.dependent_vars[0].name not in bound_arguments.arguments):
```
pckroon: Except that now the continuation line is underindented and looks like the body of the if.

tests/test_objectives.py
```diff
@@ -189,7 +195,41 @@ def test_LogLikelihood(self):
         self.assertAlmostEqual(fit_exact_result.stdev(a), fit_num_result.stdev(a))
         self.assertAlmostEqual(fit_exact_result.stdev(b), fit_num_result.stdev(b))
 
+    def test_data_sanity(self):
+        """
+        Tests very basicly the data sanity for different objective types.
+        :return:
+        """
+        # Create test data
+        xdata = np.linspace(0, 100, 25)  # From 0 to 100 in 100 steps
```
pckroon: 25 steps?

tests/test_objectives.py
```diff
@@ -189,7 +195,41 @@ def test_LogLikelihood(self):
...
+        a_vec = np.random.normal(15.0, scale=2.0, size=xdata.shape)
+        b_vec = np.random.normal(100, scale=2.0, size=xdata.shape)
+        ydata = a_vec * xdata + b_vec  # Point scattered around the line 5 * x + 105
```
pckroon: 15*x+100

symfit/core/fit.py
```diff
@@ -1337,6 +1352,55 @@ def _determine_minimizer(self):
...
+                        bound_arguments.arguments[var.name] = None
+                    else:
+                        raise TypeError(
+                            'A value was provided for `{}`, however for {} '
+                            'fitting the dependent variable cannot have a value '
+                            'assigned to it.'.format(var.name, objective)
+                        )
+            return objective
+        else:
+            if minimizer is MINPACK:
```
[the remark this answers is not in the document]
tBuLi: Yes indeed. MINPACK is inheritely a chi squared optimization, so I'm not sure what will happen if you use it for LogLikelihood but it won't be pretty.

symfit/core/fit.py
```diff
@@ -1337,6 +1352,55 @@ def _determine_minimizer(self):
...
+        :param minimizer: :class:`~symfit.core.minimizers.BaseMinimizer`
+            provided by the user, or  ``None``
+        :param bound_arguments: Instance of :class:`inspect.BoundArguments`.
+        :return: a subclass of `BaseObjective`.
+        """
+        if objective is not None:
+            # Client is king, return the clients objective. For LogLikelihood
+            # though, we do check if the data is compatible with likelihood
+            # fitting.
+            if (objective is LogLikelihood or objective is MinimizeModel or
```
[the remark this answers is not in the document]
tBuLi: The problem here is that objective could be an instance, and calling `issubclass` with an instance gives `TypeError: issubclass() arg 1 must be a class`. I'm therefore not sure what the cleanest solution is here.

symfit/core/objectives.py
```diff
@@ -367,7 +383,30 @@ def eval_hessian(self, ordered_parameters=[], **parameters):
...
+    @cached_property
+    def sigma_data(self):
+        """
+        :return: Empty OrderedDict.
+        :rtype: collections.OrderedDict
+        """
+        return OrderedDict()
+
+
+class LogLikelihood(HessianObjective, BaseIndependentObjective):
```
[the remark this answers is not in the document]
tBuLi: No, only on the independent data. For LL we only know the values for the independent variables. (The dependent one would have to be the probability for that measurement, which is what we don't know.)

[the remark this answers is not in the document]
pckroon: Huh? Doesn't LL describe the log likelihood of a dependent data being produced by a model with given parameters and independent data?
Wikipedia time for me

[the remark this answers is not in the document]
pckroon: As I read it wiki agrees with me:
> ... Intuitively, this selects the parameter values that make the data most probable.

but it's not being super explicit.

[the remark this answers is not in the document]
tBuLi: Suppose I have sampled data which (I suspect) is from a gaussian distribution, with pdf:

```python
f(x) = exp(- (x - x0)**2 / (2 * sig**2)) / sqrt(2 * pi * sig**2)
```
The data we are confronted with are the `x`, which we suspect is from this gaussian distribution. We therefore try to find the parameters `x0` and `sig` which make this dataset the most probable. But at no point do we have access to `f(x)`. This is why we have no dependent data for a likelihood problem.

So it is not the same as first creating a histogram of our `x`, which will then give us an estimate of `f(x)` within a bin of with `dx`, which we can then fit with a least squares.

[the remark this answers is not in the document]
pckroon: Aaaahh that starts to make sense! Ok then!

PR #237, opened 2019-05-07
LogLikelihood global fitting support, fixed printing of special functions.

### LINKED ISSUE #234: Global LogLikelihood fails
In principle LogLikelihood fitting should also work when fitting to multiple data sets, however this [SO ](https://stackoverflow.com/questions/56006357/global-distribution-fitting-sharing-some-parameters-without-any-specification-of)question shows that it doesn't.

The issue seems to be that numpy as not able to take the log if the model components have different sizes. We should therefore take this log per component before summing.

### LINKED ISSUE #236: Printing of special function (bessel etc.)
This SO question brought to light that special functions, like BesselJ etc, are not currently conferted to their scipy counterparts. This is because out `SymfitNumpyPrinter` inherits from `NumpyPrinter`, but special functions are a part of `SciPyPrinter`.

Therefore, we should consider switching, although the `SciPyPrinter` suffers from namespace issues because scipy itself sometimes does.

tBuLi:
Fixed #234 and Fixed #236.

Unfortunately SymPy printing has to be monkeypatched, I see no other way to solve this. Subclassing just doesn't seem to do the trick.

### DIFF
--- a/symfit/core/objectives.py
+++ b/symfit/core/objectives.py
@@ -413,11 +413,9 @@ class LogLikelihood(HessianObjective):
                         df / component
                     )
                 )
-            else:
-                result.append(component_sums)
-        else:
-            result = np.sum(result, axis=0)
-            return np.atleast_1d(np.squeeze(np.array(result)))
+            result.append(component_sums)
+        result = np.sum(result, axis=0)
+        return np.atleast_1d(np.squeeze(np.array(result)))
 
     def eval_hessian(self, ordered_parameters=[], **parameters):
         """
@@ -446,8 +444,8 @@ class LogLikelihood(HessianObjective):
             # We sum away everything except the matrices in the axes 0 & 1.
             axes = tuple(range(2, len(dd_logf.shape)))
             result += np.sum(dd_logf, axis=axes, keepdims=False)
-        else:
-            return np.atleast_2d(np.squeeze(np.array(result)))
+
+        return np.atleast_2d(np.squeeze(np.array(result)))
 
 
 class MinimizeModel(HessianObjective):
--- a/symfit/core/objectives.py
+++ b/symfit/core/objectives.py
@@ -381,7 +381,9 @@ class LogLikelihood(HessianObjective):
             ordered_parameters, **parameters
         )
 
-        ans = - np.nansum(np.log(evaluated_func))
+        ans = - np.nansum(
+            [np.nansum(np.log(component)) for component in evaluated_func]
+        )
         return ans
 
     @keywordonly(apply_func=np.nansum)
@@ -403,14 +405,18 @@ class LogLikelihood(HessianObjective):
         )
 
         result = []
-        for jac_comp in evaluated_jac:
+        for component, jac_comp in zip(evaluated_func, evaluated_jac):
+            component_sums = []
             for df in jac_comp:
-                result.append(
+                component_sums.append(
                     - apply_func(
-                        df.flatten() / evaluated_func
+                        df / component
                     )
                 )
+            else:
+                result.append(component_sums)
         else:
+            result = np.sum(result, axis=0)
             return np.atleast_1d(np.squeeze(np.array(result)))
 
     def eval_hessian(self, ordered_parameters=[], **parameters):
--- a/symfit/core/printing.py
+++ b/symfit/core/printing.py
@@ -1,3 +1,12 @@
+"""
+``symfit`` occasionally updates the printing of ``sympy`` objects, such that they
+print into their ``numpy``/``scipy`` equivalent. This is done because sometimes
+such printing has not been implemented in ``sympy`` yet, or because we want
+slightly different behavior from the standard one.
+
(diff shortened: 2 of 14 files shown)


symfit/core/objectives.py
```diff
@@ -403,14 +405,18 @@ def eval_jacobian(self, ordered_parameters=[], **parameters):
...
             for df in jac_comp:
-                result.append(
+                component_sums.append(
                     - apply_func(
-                        df.flatten() / evaluated_func
+                        df / component
                     )
                 )
+            else:
+                result.append(component_sums)
```
pckroon: This else is pointless since there's no break in the for loop.

symfit/core/objectives.py
```diff
@@ -403,14 +405,18 @@ def eval_jacobian(self, ordered_parameters=[], **parameters):
...
         else:
+            result = np.sum(result, axis=0)
```
pckroon: Same here.

symfit/core/printing.py
```diff
@@ -1,9 +1,9 @@
 from sympy.printing.pycode import NumPyPrinter
 from sympy.printing.codeprinter import CodePrinter
 
-#########################################################
-# Monkeypatch the printer until this is merged upstream #
-#########################################################
+##########################################################
+# Monkeypatch the printers until this is merged upstream #
+##########################################################
```
pckroon: It should probably be mentioned in the docs that we monkeypatch some printers which may cause strange results for users that import symfit, but expect sympy printing elsewhere in their code. Different PR.

tBuLi:
The main issue was that something like using `Gamma` was not printed to the proper scipy equivalent, because our printer inherited from `NumpyPrinter` but not from `SciPyPrinter`. I then tried changing that but that didn't work, and then I found out that the reason why is because the printer you supply to `lambdify` has to be a `LambdaPrinter`, which somehow or another has to do some work before the module printers are called. If no printer is provided, `modules=['sympy', 'numpy']`, and therefore the correct printing is done.

I also tried what is suggested here: https://docs.sympy.org/latest/modules/utilities/lambdify.html, which is to provide a mapping for custom functions along with the modules (ctrl+f "Step 2 is augmented by certain translations" on that page). But in that case I still got the wrong result, so I gave up and monkey patched it.

pckroon:
I *think* I understand. In some places we want to replace the LambdaPrinter (step 1 in the lambdify process) to create a different string to evaluate (e.g. matmul), and in other cases we want to affect the mapping from that string to numpy/scipy functions (step 2).

Step 1 we can change by creating a NumpyPrinter subclass, and step 2 we can influence by feeding the right dict to `lambdify`.
Also, I can't find a ScipyPrinter anywhere in the docs.

tBuLi:
I now remember what my problem was with giving a custom mapping to `modules` to change certain printing behavior: this dict needs to contain the function that should replace the functionality, not the change in printer method. e.g. `{'sin': numpy.sin}` will make `sin` print as a numpy `sin`, but our case we sometimes make the choice based on certain parameters. e.g. print `A**2` using a normal power if it is a 1x1 matrix, but using matrix power if it is larger.

The only way I see is to make a new function that chooses on runtime, but that gives overhead. The current implementation will print it as the correct function.

pckroon:
Right.
> we sometimes make the choice based on certain parameters. e.g. print A**2 using a normal power if it is a 1x1 matrix, but using matrix power if it is larger.

has to be done with a subclass of NumpyPrinter, which affects the string representation we end up with.
Then, for things like DiracDelta and probably Besseli, you can use the dict approach to affect how that string gets translated to code.

tBuLi:
I think I have found a good solution to these problems. The solution is to define `_numpycode` on the symbolical objects in question, so that this can be picked up by the `NumPyPrinter` of sympy.

This is the smallest intrusion we can make, because although we will have to slightly monkey patch the relevant objects, it will only affect their behavior under the `NumPyPrinter`. It will not affect their behavior under other printers, nor will we be tampering with the printers themselves. Additionally, defining this so called `printmethod` is the solution expected by sympy, since the printer will always check this attribute before trying one of the built in options.

tests/test_auto_fit.py
```diff
@@ -250,7 +250,7 @@ def test_gaussian_2d_fitting(self):
         y = Variable('y')
         g = Variable('g')
 
-        model = Model({g: A * Gaussian(x, x0, sig_x) * Gaussian(y, y0, sig_y)})
+        model = GradientModel({g: A * Gaussian(x, x0, sig_x) * Gaussian(y, y0, sig_y)})
```
pckroon: Why? And shouldn't this be another PR?

tBuLi: From my last comment:

> As a small aside, I have decided not to print the DiracDelta anymore for fear of false positives. It is much more useful that users get an error about this delta function being there, so they can reconsider what needs to be done in the first place since you probably shouldn't be having them. I therefore changed the relevant tests to use GradientModel instead of Model. I do think this might also be an issue in sympy, since mathematica does not give delta peaks for second derivatives of Gaussian functions.

PR #238, opened 2019-05-14
Lagrange mult objective

tBuLi:
Description will follow after lunch

### DIFF
--- a/symfit/core/fit.py
+++ b/symfit/core/fit.py
@@ -989,16 +989,6 @@ class TakesData(object):
             else:
                 self.absolute_sigma = False
 
-    def _make_parameters(self):
-        """
-        Make :class:`inspect.Parameter`'s using the default rules: all variables
-        to the model need to be provided.
-
-        :return: list of  :class:`inspect.Parameter` corresponding to
-            ``self.model``.
-        """
-        return self._make_signature_params_for_objective(self.model, BaseObjective)
-
     def _make_signature(self):
         """
         Make a :class:`inspect.Signature` object corresponding to
@@ -1007,31 +997,33 @@ class TakesData(object):
         :return: :class:`inspect.Signature` object corresponding to
             ``self.model``.
         """
-        parameters = self._make_parameters()
+        parameters = self._make_parameters(self.model)
         parameters = sorted(parameters, key=lambda p: p.default is None)
         return inspect_sig.Signature(parameters=parameters)
 
     @staticmethod
-    def _make_signature_params_for_objective(model, objective_cls):
+    def _make_parameters(model, none_allowed=None):
         """
-        Based on an objective function, return the inspect.Parameter objects
-        needed to satisfy this objective.
+        Based on a model, return the inspect.Parameter objects
+        needed to satisfy all the variables of this model.
 
         :param model: instance of model
-        :param objective_cls: (subclass) of objective.
+        :param none_allowed: If provided, this has to be a sequence of
+            :class:`symfit.core.argument.Variable` whose values are set to
+            ``None`` by default. If not provided, this will be set to sigma
+            variables only.
+        :return: list of  :class:`inspect.Parameter` corresponding to all the
+            external variables of the model.
         """
-        if issubclass(objective_cls, BaseIndependentObjective):
-            vars = model.independent_vars
-        else:
-            vars = model.vars
-
+        if none_allowed is None:
+            none_allowed = model.sigmas.values()
         parameters = [
             inspect_sig.Parameter(
                 var.name,
                 kind=inspect_sig.Parameter.POSITIONAL_OR_KEYWORD,
-                default=None if var in model.sigmas.values() else inspect_sig.Parameter.empty
+                default=None if var in none_allowed else inspect_sig.Parameter.empty
             )
-            for var in vars
+            for var in model.vars
         ]
         return parameters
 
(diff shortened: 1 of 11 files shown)


Jhsmit:
Are you getting the stew or the classic vol-au-vent?

tBuLi:
You guys had stew or vol-au-vent today? I had a hamburger with spinach, Popeye would approve.

symfit/core/fit.py
```diff
@@ -1236,18 +1236,25 @@ class Fit(HasCovarianceMatrix):
...
+            will instead solve for a stationary point of the Lagrangian,
+            ``stationary_point`` will therefore be forced to ``True``.
         :param bool absolute_sigma: True by default. If the sigma is only used
             for relative weights in your problem, you could consider setting it
             to False, but if your sigma are measurement errors, keep it at True.
             Note that curve_fit has this set to False by default, which is
             wrong in experimental science.
+        :param bool stationary_point: If ``True``, solve for a stationary value
+            of the objective, i.e. when the jacobian of the objective equals
+            zero, instead of minimizing the model itself. The default setting
```
pckroon: model or objective

symfit/core/fit.py
```diff
@@ -1274,6 +1282,13 @@ def __init__(self, model, *ordered_data, **named_data):
         else:
             self.model = Model(model)
 
+        if isinstance(constraints, dict):
+            # Split in multipliers and constraints, respecting order.
+            ordered_constraints = OrderedDict(**constraints)
```
pckroon: I don;t think you need to unpack a dict to turn it into an ordereddict.

symfit/core/fit.py
```diff
@@ -1476,6 +1506,59 @@ def _init_constraints(self, constraints, model):
...
+
+        Because the constraints are met at stationary points of the Lagrange
+        function, we solve :math:`\\nabla_p \\mathcal{L} = 0` instead of
+        minimizing :math:`\\mathcal{L}` directly.
+
+        :param objective: Objective :math:`f(\\vec{p})`.
+        :param multipliers: list of :class:`symfit.core.argument.Parameter`
+            objects, to be interpreted as Lagrange multipliers :math:`\\lambda_j`.
+        :param constraint_objectives: list of objectives representing the
+            constraints :math:`g_{j}(\\vec{p})`.
```
pckroon: Why not just keep these two (multipliers, constraint_objectives) as dict for the argument here?

symfit/core/fit.py
```diff
@@ -1476,6 +1506,59 @@ def _init_constraints(self, constraints, model):
...
+            grad_gs = [constraint.model.dependent_vars[0] for constraint in
+                       constraint_objectives]
+        else:
+            grad_gs = []
+
+        # Lagrangian itself
+        model_dict = {grad_f: objective.eval_jacobian}
+        model_dict.update(
+            {grad_g_i: constraint_objectives[i].eval_jacobian for i, grad_g_i in
+             enumerate(grad_gs)})
```
pckroon: closing bracket `)` on the next line

symfit/core/fit.py
```diff
@@ -1476,6 +1506,59 @@ def _init_constraints(self, constraints, model):
...
+        model_dict[grad_L] = grad_f + sum(
+            l_i * grad_g_i for l_i, grad_g_i in zip(multipliers, grad_gs))
```
pckroon: Also here

symfit/core/fit.py
```diff
@@ -1476,6 +1506,59 @@ def _init_constraints(self, constraints, model):
...
+
+        connectivity_mapping = {grad_f: set(objective.model.free_params)}
+        connectivity_mapping.update(
+            {grad_g_i: set(objective.model.free_params) for grad_g_i in
+             grad_gs})
```
pckroon: And here

symfit/core/fit.py
```diff
@@ -1476,6 +1506,59 @@ def _init_constraints(self, constraints, model):
...
+        connectivity_mapping.update({grad_L: {grad_f, *grad_gs}})
+
+        grad_lagrangian = CallableNumericalModel(model_dict,
```
pckroon: Why numerical? Don't you (also) have  an analytical form of the lagrangian?

tests/test_auto_fit.py
```diff
@@ -304,5 +305,70 @@ def test_gaussian_2d_fitting_background(self):
...
+        """
+        x, y = variables('x, y')
+        a, b = parameters('a, b')
+        model = Model({y: a * x + b})
+
+        xdata = np.linspace(-10, 10)
+        ydata = model(x=xdata, a=2, b=3).y
+
+        # Because this model falls of the infinity, a normal minimizer wont work
+        # This is especially true when bounds are provided.
```
pckroon: Comment is not correct for this test?

symfit/core/fit.py
tBuLi: This is a key issue I'm having with this PR: this is indeed defined best for problems which have an analytical form. However, I'm encountering two problems with this. Apart from the obvious problem that not every model is analytical, the bigger issue is that something like our LeastSquared solver is not analytical, but the lagrange multiplier should be added to this LeastSquare, not to the model. That's a serious problem.

One solution that would keep this code very clean might be to say that this is only valid for cases where we need to `MinimizeModel` instead, since there adding the multipliers is trivial. But that would mean that at least for now it will not work for least squares fitting with data, unless the user writes out the least squares function symbolically, which is completely fair game and would work. But I don't want to do that out of the box because we cannot know how many dimension we might have to sum over etc.

pckroon: I think for a first implementation it would be acceptable to limit it to `MinimizeModel`. We'll need to think about what to do with the other Objectives.

PR #242, opened 2019-06-02
Notebooks in documentation

tBuLi:
Support notebooks in our documentation. This makes writing examples a lot simpler and more beautiful.


.travis.yml
```diff
@@ -41,8 +41,12 @@ jobs:
 
     - stage: docs
       python: "3.7"
+      before_install:
+        - sudo apt-get install pandoc
       install:
         - pip install --upgrade sphinx
+        - pip install --upgrade nbsphinx
+        - pip install --upgrade IPython
```
pckroon: Could you make a `requirements-docs.txt` instead of cramming it all here? That way we can do `pip install -U -r requirements-docs.txt`. That way it's also clear *why* those packages need to be installed, and it helps third parties to find all dependencies.

pckroon:
What I don't like about this is that it's much harder to review the changes to the ipynb files, but I'll live with it.
And please leave the plain .py files in /examples --- they're easier to find and quickly glance over IMHO.

tBuLi:
I agree that notebooks really suck when you try to diff them. A workaround would be to view them directly using github, like [here](https://github.com/tBuLi/symfit/blob/sphinx_notebook/docs/examples/ex_bivariate_likelihood.ipynb). However, still not a diff of course, but readable at least.

But I do think the advantages are good selling points, they're pretty much self testing, and they are easy and quick to write which lowers the threshold to writing examples, not to mention a lot of users probably work in notebooks.

.travis.yml
```diff
@@ -44,9 +44,7 @@ jobs:
       before_install:
         - sudo apt-get install pandoc
```
pckroon: There is a better way of doing this: https://docs.travis-ci.com/user/installing-dependencies/#adding-apt-packages

requirements_docs.txt
```diff
@@ -0,0 +1,10 @@
+numpy >= 1.12
+scipy >= 1.0, <1.2
+sympy >= 1.2
+toposort
+funcsigs; python_version < '3.0'
+functools32; python_version < '3.0'
```
pckroon: Can't you say that this file should include the normal requirements.txt?

tBuLi: I guess we can: https://stackoverflow.com/questions/17803829/how-to-customize-a-requirements-txt-for-multiple-environments

Issue #143, opened 2018-03-16
Grid example missing in docs


tBuLi:
An example of fitting a function of two parameters should be added to the docs. This test code could be used: https://github.com/tBuLi/symfit/blob/4ef619491a9979e1e3371047b23ba9739fa4207d/tests/test_auto_fit.py#L225

Or this github answer:
https://stackoverflow.com/questions/47874635/fitting-a-polynomial-surface-to-a-3d-data-set/49317230#49317230

PR #245, opened 2019-06-10
Add paragraph "Fitting moredimensional datasets" with an example

DuckNrOne:
Add paragraph "Fitting moredimensional datasets" with an example from [StackOverflow](https://stackoverflow.com/a/49317230/11008918). 
Solving Issue [#143](https://github.com/tBuLi/symfit/issues/143)
I corrected the formatting since the last pull request.

### DIFF
--- a/docs/fitting_types.rst
+++ b/docs/fitting_types.rst
@@ -386,17 +386,22 @@ are added squared, ready to be minimized. Unlike in the above example, the
   Do not cite the overall :math:`R^2` given by :mod:`symfit`.
 
 Fitting multidimensional datasets
---------------------------------
-Imagine that you have now a specific property over a grid, like the surface height,
-so you have a three-dimensional data set. To fit a function with two parameters
-to this data set, you can use :mod:`symfit`.
+----------------------------------
+So far we had only considered one variable, but in the real world, we have
+problems with more mutable variables, for example, a specific property over a
+grid, like a temperature. In this case, you have a three-dimensional data set
+(x-, y-coordinates, temperature).
 
-Let's work this out with the following example. We made a polynomial function
-with two coefficients, representing two terms of mixed order in ``x`` and ``y``
+If we want to fit a function with two parameters (x- and y-coordinates) to
+this datasets, we can use :mod:`symfit`.
 
-:math:`f(x, y) = z = c2 \cdot x^4 \cdot y^5 + c1 \cdot x \cdot y^2`
+Let's work this out with the following mathematical model. We have a polynomial
+function with two coefficients, representing two terms of mixed order in ``x``
+and ``y``
 
-First, we have to define our model::
+:math:`T(x, y) = z = c2 \cdot x^4 \cdot y^5 + c1 \cdot x \cdot y^2`
+
+Secondly, we have to implement our  model::
 
   x, y, z = variables('x, y, z')
   c1, c2 = parameters('c1, c2')
@@ -409,8 +414,9 @@ Now I can fit this polynomial model to some data::
   fit = Fit(model, x=xdata, y=ydata, z=zdata)
   fit_result = fit.execute()
 
-In conclusion, we made a polynomial function with two parameters,
-which coordinates of a mesh could be.
+In conclusion, we made a mathematical model for a multidimensional function
+with two parameters, implemented this model and feet it this data
+to get a result.
 
 Global Minimization
 -------------------
--- a/docs/fitting_types.rst
+++ b/docs/fitting_types.rst
@@ -392,7 +392,7 @@ so you have a three-dimensional data set. To fit a function with two parameters
 to this data set, you can use :mod:`symfit`.
 
 At this example, we made a polynomial function with two coefficients,
-representing two terms of mixed order in :mod:`x` and :mod:`y`::
+representing two terms of mixed order in ``x`` and ``y``::
 
   x, y, z = variables('x, y, z')
   c1, c2 = parameters('c1, c2')
--- a/docs/fitting_types.rst
+++ b/docs/fitting_types.rst
@@ -397,9 +397,9 @@ this datasets, we can use :mod:`symfit`.
 
 Let's work this out with the following mathematical model. We have a polynomial
 function with two coefficients, representing two terms of mixed order in ``x``
-and ``y``
+and ``y``.
 
-:math:`T(x, y) = z = c2 \cdot x^4 \cdot y^5 + c1 \cdot x \cdot y^2`
+:math:`T(x, y) = z = c_2 x^4 y^5 + c_1 x y^2`
 
 Secondly, we have to implement our  model::
 
@@ -407,9 +407,15 @@ Secondly, we have to implement our  model::
   c1, c2 = parameters('c1, c2')
   model_dict = {z: Poly( {(1, 2): c1, (4, 5): c2}, x ,y)}
   model = Model(model_dict)
-  #printed version: z(x, y; c1, c2) = Poly(c2*x**4*y**5 + c1*x*y**2, x, y, domain='ZZ[c1,c2]')
+  # prints z(x, y; c1, c2) = Poly(c2*x**4*y**5 + c1*x*y**2, x, y, domain='ZZ[c1,c2]')
 
-Now I can fit this polynomial model to some data::
+Now we can fit this polynomial model to some data, where ``xdata``, ``ydata``
+and ``zdata`` are two-dimensional::
+
+  x = np.linspace(-10, 10, 0.5)
+  y = np.linspace(-10, 10, 0.5)
+  xdata, ydata = np.meshgrid(x, y)
+  zdata = 42 * xdata**4 * ydata**5 + 3.14 * xdata * ydata**2
 
   fit = Fit(model, x=xdata, y=ydata, z=zdata)
   fit_result = fit.execute()
--- a/docs/fitting_types.rst
+++ b/docs/fitting_types.rst
@@ -385,28 +385,27 @@ are added squared, ready to be minimized. Unlike in the above example, the
 
   Do not cite the overall :math:`R^2` given by :mod:`symfit`.
 
-  Fitting moredimensional datasets
-  --------------------------------
+Fitting moredimensional datasets
+--------------------------------
+Imagine that you have now a specific property over a grid, like a temperature,
+so you have a three-dimensional data set. To fit a function with two parameters
+to this data set, you can use :mod:`symfit`.
 
-  Imagine that you have now a specific property over a grid, like a temperature,
-  so you have a three-dimensional data set. To fit a function with two parameters
-  to this data set, you can use :mod:`symfit`.
+At this example, we made a polynomial function with two coefficients,
+representing two terms of mixed order in x and y::
 
-  At this example, we made a polynomial function with two coefficients,
-  representing two terms of mixed order in x and y::
+  x, y, z = variables('x, y, z')
+  c1, c2 = parameters('c1, c2')
+  model_dict = {z: Poly( {(1, 2): c1, (4, 5): c2}, x ,y)}
+  model = Model(model_dict)
 
-    x, y, z = variables('x, y, z')
-    c1, c2 = parameters('c1, c2')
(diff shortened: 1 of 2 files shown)


docs/fitting_types.rst
```diff
@@ -385,6 +385,28 @@ are added squared, ready to be minimized. Unlike in the above example, the
 
   Do not cite the overall :math:`R^2` given by :mod:`symfit`.
 
+Fitting moredimensional datasets
+--------------------------------
+Imagine that you have now a specific property over a grid, like a temperature,
+so you have a three-dimensional data set. To fit a function with two parameters
+to this data set, you can use :mod:`symfit`.
+
+At this example, we made a polynomial function with two coefficients,
```
pckroon: "At this example" should be "For this example". I think. I'm not a native english speaker.

docs/fitting_types.rst
```diff
@@ -385,6 +385,28 @@ are added squared, ready to be minimized. Unlike in the above example, the
...
+representing two terms of mixed order in x and y::
```
pckroon: x and y probably need some formatting. I'm not sure what the syntax is at the moment though.

docs/fitting_types.rst
```diff
@@ -385,6 +385,28 @@ are added squared, ready to be minimized. Unlike in the above example, the
...
+
+  x, y, z = variables('x, y, z')
+  c1, c2 = parameters('c1, c2')
+  model_dict = {z: Poly( {(1, 2): c1, (4, 5): c2}, x ,y)}
```
pckroon: Most readers won't know this `Poly` syntax. It would be good to either print the model made, or write out the polynomial above.

docs/fitting_types.rst
DuckNrOne: I checked it with grammarly. "At this example" should be correct.

docs/fitting_types.rst
DuckNrOne: Do you mean something like a link or bold text?

pckroon: Italics or even better, mathtyped. I'm also not sure whether we did it in the rest of the docs.

DuckNrOne: Italic is no problem, but I don't know the syntax of mathtyped text. Moreover I didn't notice any formatted text for variables in the docs. But this could be a new issue to work on ;)

docs/fitting_types.rst
DuckNrOne: I would add the following line at the end of the first codeblock:
`#printed version: z(x, y; c1, c2) = Poly(c2*x**4*y**5 + c1*x*y**2, x, y, domain='ZZ[c1,c2]')`

pckroon: Or:
```python3
# >>> print(model)
# z(x, y; c1, c2) = Poly(c2*x**4*y**5 + c1*x*y**2, x, y, domain='ZZ[c1,c2]')
```
Or put the mathematical form (:math:`c2*x^4*y^5 + c1*x*y^2`) near where you say "we made a polynomial function"

DuckNrOne: I'll add the mathematical form

tBuLi: I would keep the `Poly` object because it is far more powerful since it can generate arbitrary polynomials. I think it would be nice if you could elaborate a bit on the math form. If you write using `:math:` in the docs which polynomial you are making, then it will also be clearer in the code even without a print.

docs/fitting_types.rst
```diff
@@ -385,6 +385,29 @@ are added squared, ready to be minimized. Unlike in the above example, the
 
   Do not cite the overall :math:`R^2` given by :mod:`symfit`.
 
+Fitting moredimensional datasets
```
tBuLi: The mathematical term is multidimensional instead of moredimensional

pckroon:
Looks good to me, thanks :)

docs/fitting_types.rst
tBuLi: I'm definitely disagreeing with grammarly then ;). I would say 'in this example' or even better: 'As an example we made a polynomial function with two coefficients [...]'

docs/fitting_types.rst
DuckNrOne: I changed this with the last push

tBuLi:
Thanks indeed for helping improve the documentation. It's much appreciated :). 

As for the general style of your example, I would prefer it if you first introduce the concept of a multidimensional problem in general, to convey the general idea that we are talking about functions of type `:math: z(x_1, \ldots, x_n)`, such as the temperature distribution on a plane `:math:T(x, y)` as you had originally.

Then you can move into your polynomial example, introducing `:math: z(x, y) = ...`, and giving the code example of how to fit to this.

docs/fitting_types.rst
```diff
@@ -385,6 +385,29 @@ are added squared, ready to be minimized. Unlike in the above example, the
...
+--------------------------------
+Imagine that you have now a specific property over a grid, like the surface height,
+so you have a three-dimensional data set. To fit a function with two parameters
+to this data set, you can use :mod:`symfit`.
+
+At this example, we made a polynomial function with two coefficients,
+representing two terms of mixed order in ``x`` and ``y``::
```
tBuLi: Instead of this `::`, you can use a `.. math::` block to first write the equation using latex. With that example in mind, the Poly syntax is more understandable.

DuckNrOne: The version you used is not the newest. I already implemented the mathematical form. 
Please correct me, if I'm wrong :)

docs/fitting_types.rst
```diff
@@ -385,6 +385,33 @@ are added squared, ready to be minimized. Unlike in the above example, the
...
+Fitting multidimensional datasets
+--------------------------------
+Imagine that you have now a specific property over a grid, like the surface height,
+so you have a three-dimensional data set. To fit a function with two parameters
+to this data set, you can use :mod:`symfit`.
+
+Let's work this out with the following example. We made a polynomial function
+with two coefficients, representing two terms of mixed order in ``x`` and ``y``
+
+:math:`f(x, y) = z = c2 \cdot x^4 \cdot y^5 + c1 \cdot x \cdot y^2`
```
tBuLi: :math:`z = f(x, y) = c_2 x^4 y^5 + c_1 x y^2`

for proper latex typesetting. The `\cdot` are of course optional, but personally I would prefer not to write them.

docs/fitting_types.rst
```diff
@@ -385,6 +385,33 @@ are added squared, ready to be minimized. Unlike in the above example, the
...
   Do not cite the overall :math:`R^2` given by :mod:`symfit`.
 
+Fitting multidimensional datasets
+--------------------------------
+Imagine that you have now a specific property over a grid, like the surface height,
+so you have a three-dimensional data set. To fit a function with two parameters
+to this data set, you can use :mod:`symfit`.
+
+Let's work this out with the following example. We made a polynomial function
+with two coefficients, representing two terms of mixed order in ``x`` and ``y``
```
tBuLi: Forgot a `:` at the end of the sentence.

docs/fitting_types.rst
```diff
@@ -385,6 +385,33 @@ are added squared, ready to be minimized. Unlike in the above example, the
...
+
+:math:`f(x, y) = z = c2 \cdot x^4 \cdot y^5 + c1 \cdot x \cdot y^2`
+
+First, we have to define our model::
+
+  x, y, z = variables('x, y, z')
+  c1, c2 = parameters('c1, c2')
+  model_dict = {z: Poly( {(1, 2): c1, (4, 5): c2}, x ,y)}
+  model = Model(model_dict)
+  #printed version: z(x, y; c1, c2) = Poly(c2*x**4*y**5 + c1*x*y**2, x, y, domain='ZZ[c1,c2]')
```
tBuLi: Perhaps change this to `# prints z(x, y; c1, c2) = Poly(c2*x**4*y**5 + c1*x*y**2, x, y, domain='ZZ[c1,c2]')`, or actually print it in a seperate codeblock like 
```python
>>> print(model)
z(x, y; c1, c2) = Poly(c2*x**4*y**5 + c1*x*y**2, x, y, domain='ZZ[c1,c2]')
```

docs/fitting_types.rst
```diff
@@ -385,6 +385,33 @@ are added squared, ready to be minimized. Unlike in the above example, the
...
+
+Now I can fit this polynomial model to some data::
+
+  fit = Fit(model, x=xdata, y=ydata, z=zdata)
+  fit_result = fit.execute()
+
+In conclusion, we made a polynomial function with two parameters,
+which coordinates of a mesh could be.
```
tBuLi: Actually it is a function of both two parameters and two variables. The variables span the mesh (x, y), and we obtained two parameters.

tBuLi: Actually we should probably also comment that xdata and ydata here are two dimensional, not one dimensional.

I think you should actually generate some mock data here using `np.meshgrid` to really make this example clear.

docs/fitting_types.rst
```diff
@@ -385,6 +385,33 @@ are added squared, ready to be minimized. Unlike in the above example, the
...
+
+First, we have to define our model::
+
+  x, y, z = variables('x, y, z')
+  c1, c2 = parameters('c1, c2')
+  model_dict = {z: Poly( {(1, 2): c1, (4, 5): c2}, x ,y)}
+  model = Model(model_dict)
+  #printed version: z(x, y; c1, c2) = Poly(c2*x**4*y**5 + c1*x*y**2, x, y, domain='ZZ[c1,c2]')
+
+Now I can fit this polynomial model to some data::
```
tBuLi: You switched from `we` to `I`.

docs/fitting_types.rst
DuckNrOne: Done

docs/fitting_types.rst
DuckNrOne: Done :D

docs/fitting_types.rst
DuckNrOne: Done

docs/fitting_types.rst
DuckNrOne: I would take the commented

docs/fitting_types.rst
DuckNrOne: > Actually it is a function of both two parameters and two variables. The variables span the mesh (x, y), and we obtained two parameters.

I improved the text bevor I read your comment. Fits that as it is now?

DuckNrOne: > Actually we should probably also comment that xdata and ydata here are two dimensional, not one dimensional.
> 
> I think you should actually generate some mock data here using `np.meshgrid` to really make this example clear.

Just to be clear: xdata, ydata and zdata are two dimensional, right? :D

docs/fitting_types.rst
tBuLi: Alright I have no problem with that, but please do include a space after the # to make it pep 8 compliant

DuckNrOne: Done 👍

docs/fitting_types.rst
tBuLi: yes indeed, all have the same shape :).

DuckNrOne: I would add something like this to the second code block:

`x = np.arange(-10, 10, 0.4)`
 `y = np.arange(-10, 10, 0.4)`
 `xdata, ydata = np.meshgrid(x, y)`
 `zdata = 42 * xdata**4 * ydata**5 + 3.14 * xdata * ydata**2`

tBuLi: Yeah, I like that, especially your choice of parameter values ;).

Small change: with non integer steps it is better to use `np.linspace`.

DuckNrOne:
What else?

DuckNrOne:
Sometimes I don't get the error of Travis CI...can smb help me with this topic pls :D

tBuLi:
Try building the docs localy first with the same command we use on travis:
``sphinx-build -nW -b html docs docs_build``

That should give you some useful error messages. Usually it is mistakes with indentation and things like that.

docs/fitting_types.rst
DuckNrOne: This problem is done, right? :D

Issue #246, opened 2019-06-12
Overview of code structure needed


DuckNrOne:
Hello,

It's more a question than an issue. So if somebody wants to contribute to symfit, it's quite difficult to get an overview of the code structure. 

I know it's a big task, but is it possible to create a manual, which help to get into the code architecture?

pckroon:
https://symfit.readthedocs.io/en/stable/api_structure.html
Beyond that we probably need to improve the general structure (fit.py is too large). But if you have specific questions feel free to ask.

tBuLi:
I agree with you that the documentation is currently a bit lacking in outlining these things for people who want to contribute, so it's not unfair to ask. The link @pckroon provided is indeed the best we currently have, and even it seems a bit outdated and in need of review. So in fact, I will reopen this as an issue that needs to be addressed.

As to what we can do about this immediately, @pckroon and I are usually quite quick to respond to specific problems, so if you have something you would like to work on just let us know.

Do you have something specific you would like symfit to be able to do and would like to discuss? Perhaps you can open an issue for that and then we can discuss it there. Let me know what it is you would like to contribute :).

And if you would like me to recommend some valued issues which might be relatively simple to address, you can have a look at #160. (Not very easy but for sure highly valued by some users.)

DuckNrOne:
Thanks for your quick response. 
I already read the API structure from the docs yesterday...That's why I opened this issue :)
I want to help on this project, because I think that it has a great perspective. 

I'll have an eye on the pull request #247. (After some lines in the docs (#204 & #143)) I would try to help at the issue #160 .

DuckNrOne:
As I am currently engaged in unittesting in my free time, I would suggest that I develop more test cases for symfit. In this case, it will be possible for me to deal with the code structure further.

tBuLi:
That would in fact be very much appreciated, adding and organizing tests is something that needs to be done, and it's a good way to get to know a project!

You can have a look here to see the areas which need attention: https://coveralls.io/github/tBuLi/symfit

You can ignore the worst 2 files though, keywordonly.py and leastsqbounds.py. The first will be fixed by discontinuing py27 support, and the latter by removing that file altogether and using `scipy`'s equivalent.

pckroon:
More tests would be great indeed! We intend to move to (parametrized) pytest tests and fixtures [1] rather than the jumble it is at the moment. You are more than welcome to either add tests in this new style, convert old/existing tests or just add new test cases in the current style.
[...]

PR #248, opened 2019-06-16
Add the example "ODEModels as subproblems using CallableNumericalModel" (Issue #204)

### LINKED ISSUE #204: Fitting ode with data not directly linked to an ode variable
Hi,

Thank you for your great framework.

I'm trying to use ODE fitting when my data are not directly linked to an ode variable:
+ y : variable of an ode model with some kind of parameters to optimize
+ z_data = f(y,params) 
    where:
            * z_data : my data
            * f is a given function but it's not an ode 
            * params are additionnal parameters to optimize

After some kind of test, I didn't figure out how to do that.

If it's not possible, could you add this features?
If it's possible, could you add the procedure in the documentation because I didn't find the obvious way to do it?

DuckNrOne:
I renamed the first example about the CallableNumerical Model and added a second example about ODEModels as subproblems using CallableNumericalModel. 

This pull request fixes #204


tBuLi:
Thanks for making this PR! The notebooks are a bit hard to review so I'll do that here:

There have been some changes to the (`CallableNumericalModel`) syntax which will make it even better, I'm sorry I should've mentioned this in #204.

Change
```python
ode_model = ODEModel({D(y, x): k * y}, initial={x: 0.0, y: 1.0})
model = CallableNumericalModel({z: fun(ode_model)},independent_vars=[x], params=[k,k2])
```
To
```python
w = Variable('w')
ode_model = ODEModel({D(y, x): k * y}, initial={x: 0.0, y: 1.0})
model_dict = {
    z: 2 * w + k2,
    w: ode_model(x=x, k=k)
}
model = CallableNumericalModel(model_dict, connectivity_mapping={z: {w, k2}, w: {x, k}})
```

This is a new syntax for initiating `CallableNumericalModel` which wasn't possible at the time this issue first arose. This means the wrapping function can be removed entirely. And the `independent_vars` and `params` keywords to `CallableNumericalModel` have been deprecated in favor of `connectivity_mapping`, I think from this example you can see why :).

Additionally, remove the extra space in your prints 'k = ', print will add a space automatically.

Finally, naming-wise I would prefer to name `k2` something else, like `b`. `k` to me sounds like a rate constant, whereas `b` is traditionally the constant term in `a x + b`.

I apologize for not pointing this out earlier.

DuckNrOne:
Everything is fine :D It's not a big problem to change the code a bit. Unfortunately, I didn't find information about the ``CallableNumericalModel``, but this is another issue :D

I changed the printing thing and the naming of the parameters, but with I got an error by creating the second entry of ``model_dict``: _cannot determine truth value of Relational_. Do you have an idea?

tBuLi:
Hmm weird, using the exact code I gave above? Perhaps you can post your new code and the full traceback?

DuckNrOne:
The code: 
```python
from symfit import variables, Parameter, Fit, D, ODEModel, CallableNumericalModel
import numpy as np
import matplotlib.pyplot as plt

# Create data
x_data = np.linspace(0.0, 10.0, 1000)
a_expected = 0.6
b_expected = 10.0
y_data = 2 * np.exp(a_expected * x_data) + b_expected

# Initialise variables and parameters
w, x, y, z = variables('w, x, y, z')
a = Parameter('a', 0.0)
b = Parameter('b', 0.0)

# Define model
ode_model = ODEModel({D(y, x): a * y}, initial={x: 0.0, y: 1.0})
model_dict = {
    z: 2 * w + b,
    w: ode_model(x=x, a=a),
}
model = CallableNumericalModel(model_dict, connectivity_mapping={z: {w, b}, w: {x, a}})

# Apply model
fit = Fit(model,x=x_data, z=y_data)
fit_result = fit.execute()
```
And the Traceback:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-11-dbc1f7a64940> in <module>
     18 model_dict = {
     19     z: 2 * w + b,
---> 20     w: ode_model(x=x, a=a),
     21 }
     22 model = CallableNumericalModel(model_dict, connectivity_mapping={z: {w, b}, w: {x, a}})

~\Anaconda3\lib\site-packages\symfit\core\fit.py in __call__(self, *args, **kwargs)
   1699         """
   1700         Ans = variabletuple('Ans', self)
-> 1701         ans = Ans(*self.eval_components(*args, **kwargs))
   1702         return ans
   1703 

~\Anaconda3\lib\site-packages\symfit\core\fit.py in eval_components(self, *args, **kwargs)
   1646         else:
   1647             t_bigger = np.concatenate(
-> 1648                 (np.array([t_initial]), t_like[t_like > t_initial])
   1649             )
   1650             t_smaller = np.concatenate(

~\Anaconda3\lib\site-packages\sympy\core\relational.py in __nonzero__(self)
    227 
    228     def __nonzero__(self):
--> 229         raise TypeError("cannot determine truth value of Relational")
    230 
    231     __bool__ = __nonzero__

TypeError: cannot determine truth value of Relational
```

tBuLi:
Ah yes, I forgot some important details. The line for `w` should be
```python
w: lambda x, a: ode_model(x=x, a=a).y,
```
since models return an `Ans` object which is essentially a namedtuple with all the evaluated components. So we have to select the `y` component of the answer.

It would actually be nice if we recognize when a component returns an `Ans` so we could write 
```python
model_dict = {
    z: 2 * w + b,
    w: ode_model,
}
```
instead, since the lamba just to do the unpacking feels a bit redundant. But that should be a separate issue.

tBuLi:
Thinking about it some more, `w` is redundant, we can safely use `y` to represent the output of the ODE. That is much clearer:

```python
from symfit import variables, Parameter, Fit, D, ODEModel, CallableNumericalModel
import numpy as np
import matplotlib.pyplot as plt

# Create data
x_data = np.linspace(0.0, 10.0, 1000)
a_expected = 0.6
b_expected = 10.0
z_data = 2 * np.exp(a_expected * x_data) + b_expected

# Initialise variables and parameters
a = Parameter('a', 0.0)
b = Parameter('b', 0.0)

# Define model
ode_model = ODEModel({D(y, x): a * y}, initial={x: 0.0, y: 1.0})
model_dict = {
    z: 2 * y + b,
    y: lambda x, a: ode_model(x=x, a=a).y,
}
model = CallableNumericalModel(model_dict, connectivity_mapping={z: {y, b}, y: {x, a}})

# Apply model
fit = Fit(model, x=x_data, z=z_data)
fit_result = fit.execute()
```

DuckNrOne:
I added the new syntax now. I think it is time to look into the real code of symfit :D

@tBuLi could you please look over the description of the example :)

tBuLi:
The code looks very good right now. The text could use an additional read-through, because there is for example a space missing somewhere after a period. After you give it another read through to see if the text flows well, I can merge it. And let me know in #246 what you would like to work on in the core specifically :).

DuckNrOne:
All right. I tried to improve the text to make the example clearer and the text more fluid.

PR #250, opened 2019-06-18
Connectivity infer

tBuLi:
In the case of a CallableNumericalModel with mixed analytical and non-analytical components, the connectivity mapping corresponding to the analytical components can just be infered automatically. This is a lot easier on the eyes:

```python
a, b = parameters('a, b')
model_dict = {z: lambda y, a, b: a * y + b,
              y: x ** a}

mixed_model = CallableNumericalModel(
    model_dict, connectivity_mapping={z: {y, a, b}}
)
```
Giving the full connectivity mapping however, is also allowed.

### DIFF
--- a/symfit/core/models.py
+++ b/symfit/core/models.py
@@ -510,9 +510,10 @@ class BaseNumericalModel(BaseModel):
                     sub_model[var] = expr
             if sub_model:
                 sub_model = BaseModel(sub_model)
-                # Update with the users input. In case of conflict, this prioritizes
-                # the info given by the user.
-                connectivity_mapping.update(sub_model.connectivity_mapping)
+                # Update with the users input. In case of conflict, this
+                # prioritizes the info given by the user.
+                sub_model.connectivity_mapping.update(connectivity_mapping)
+                connectivity_mapping = sub_model.connectivity_mapping
 
             self.connectivity_mapping = connectivity_mapping
         else:
--- a/symfit/core/models.py
+++ b/symfit/core/models.py
@@ -508,10 +508,11 @@ class BaseNumericalModel(BaseModel):
             for var, expr in model.items():
                 if isinstance(expr, sympy.Basic):
                     sub_model[var] = expr
-            sub_model = BaseModel(sub_model)
-            # Update with the users input. In case of conflict, this prioritizes
-            # the info given by the user.
-            connectivity_mapping.update(sub_model.connectivity_mapping)
+            if sub_model:
+                sub_model = BaseModel(sub_model)
+                # Update with the users input. In case of conflict, this prioritizes
+                # the info given by the user.
+                connectivity_mapping.update(sub_model.connectivity_mapping)
 
             self.connectivity_mapping = connectivity_mapping
         else:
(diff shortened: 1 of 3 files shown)


symfit/core/models.py
```diff
@@ -496,10 +499,20 @@ def __init__(self, model, independent_vars=None, params=None, **kwargs):
             self.connectivity_mapping = {var: set(independent_vars + params)
                                          for var in model}
         elif connectivity_mapping:
-            self.connectivity_mapping = connectivity_mapping
             if not isinstance(model, Mapping):
                 raise TypeError('Please provide the model as a mapping, '
                                 'corresponding to `connectivity_mapping`.')
+            elif connectivity_mapping.keys() != model.keys():
```
pckroon: I feel like it's cleaner to just *always* create an initial connectivity mapping for the symbolic components, and then update it with what's been provided. So no if statements involved.

symfit/core/models.py
```diff
@@ -496,10 +499,20 @@ def __init__(self, model, independent_vars=None, params=None, **kwargs):
...
+                # Infer the connectivity mapping corresponding to the symbolical
+                # part automatically
+                sub_model = {}
+                for var, expr in model.items():
+                    if isinstance(expr, sympy.Basic):
+                        sub_model[var] = expr
+                sub_model = BaseModel(sub_model)
+                connectivity_mapping.update(sub_model.connectivity_mapping)
```
pckroon: I have a feeling this update is not good enough when there are interdependent components involved. Can't come up with an example at the moment though...

symfit/core/models.py
tBuLi: Fair point, in case the user provides (wrongful) information for the analytical components they will still get the error that way.

symfit/core/models.py
tBuLi: I cannot guarantee that it is not possible, but I have a feeling it would be very hard and luckily the error messages will be very helpful in this case: something like you forgot to give a value to this keyword argument. So it should be easily debug-able. I don't think it can lead to fundamental conflicts though, since the user always has the final say with the update happening last.

pckroon: The comment is wrong. User provided connectivity_mapping gets overwritten by sub_model connectivity_mapping. I do like the case where user input is prioritized over auto-generated data.

symfit/core/models.py
```diff
@@ -510,9 +510,10 @@ def __init__(self, model, independent_vars=None, params=None, **kwargs):
...
                 sub_model = BaseModel(sub_model)
-                # Update with the users input. In case of conflict, this prioritizes
-                # the info given by the user.
-                connectivity_mapping.update(sub_model.connectivity_mapping)
+                # Update with the users input. In case of conflict, this
+                # prioritizes the info given by the user.
+                sub_model.connectivity_mapping.update(connectivity_mapping)
+                connectivity_mapping = sub_model.connectivity_mapping
 
             self.connectivity_mapping = connectivity_mapping
```
pckroon: Now you get a funny case where sometimes (no analytical component) `self.connectivity_mapping` is an alias to the user input, and other times it isn't. So maybe add a `.copy()`. And yes, I'm splitting hairs here.

tBuLi: Yes, but it's a good remark because that is indeed a bit inconsistent. I can definitely see @Jhsmit come up with some way to complain about this in the future.

PR #259, opened 2019-06-21
added try/except to version finding

Jhsmit:
For those who use `conda develop` to use one of symfit's branches or the current master. Currently, when doing so, the version cannot be found and an error is raised.


Jhsmit:
bump @tBuLi

Issue #255, opened 2019-06-19
Interactive guess in Jupyter notebook crashes


egpbos:
I'm trying to use interactive guessing inside a Jupyter notebook (on macOS). I got it to work with the notebook backend, i.e. by first running

```python
%matplotlib notebook
```

With the default (`inline`) backend it obviously doesn't work, because it isn't interactive. I also had issues with other backends, e.g. using the TkAgg backend (which was activated automatically when I run just `%matplotlib` without explicit argument) actually crashes the Python kernel.

One suggestion that I would have is to try to implement all this using Jupyter widgets instead of Matplotlib sliders. This would also open the door to using different plotting packages for the preview, for instance `ipyvolume` for 3D previews.

[`InteractiveGuess` is also a bit poorly documented](https://symfit.readthedocs.io/en/stable/module_docs.html#contrib). One crucial sentence to add I think would be that you have to initialize it with the same parameters as a `Fit`. Currently, this is not clear.

Sorry for sounding maybe overly critical, let me just make clear that I love that interactive guessing is even available. I'm currently involved in a project where this may be necessary at some point. If that turns out to be the case, I could make a PR to work on the above mentioned issues, if you are open to that.

tBuLi:
You are right in pointing out that interactive fit probably doesn't play nice with notebooks, it was designed before we really knew or cared about notebooks. Any improvements to it are very welcome, and if you have suggestions on how we could change the design are welcome. Based on what you just pointed out, perhaps InteractiveGuess should just accept a `Fit` object as an argument, and nothing else, to prevent the problem you mentioned?

pckroon:
I (original author of the interactive guess thingy) agree with this on most points. The only problem I have with using ipywidgets everywhere is that it will then *only* in ipy notebooks, while now it only needs matplotlib which is still more commonly available.
It would be great if we/you could separate the drawing from the logic, and create a version that works with both sliders and ipy widgets. I unfortunately don't have the time to do that. 
I have played around with interactive 3D plots in matplotlib, but I don't think they are necessarily easier to interpret (but play around, maybe it works for your problem).

Any way, making InteractiveGuess take a Fit sounds like a great idea by the way. Feel free to open a PR!

egpbos:
Agreed, it's nice to also have it working without notebooks. Ok, if I get the time within my (quite short) project, I'll check it out, this would be somewhere in the coming month. If not, maybe some other time :) And I agree, making it take a Fit would be nicer.

PR #261, opened 2019-07-01
Interactive notebook

Jhsmit:
I needed this for my own stuff so I thought i'd make PR for communism. 
This implements interative guesses in jupyter notebooks, only for 1D atm.
Requires Ipywidgets
Not compatible with the contrib module interactive_guess because of the `plt.ioff()` which gets called on import. (also on just import of jupyter contrib module)


Jhsmit:
towards #255

symfit/contrib/interactive_guess/interactive_guess.py
```diff
@@ -10,7 +10,7 @@
 import numpy as np
 from scipy.stats import gaussian_kde
 
-plt.ioff()
+#plt.ioff()
```
pckroon: Is this comment still needed? And do I actually need to do the `plt.ioff()` to make it work in a normal Python script (not notebook)?

Jhsmit: Its needed to be commented out for my notebook implementation to work. I'm not sure if you really need it in a normal python file. I can give it a try if I have time.

Jhsmit:
Yes I agree that merging to an API would be a good idea and would eliminate duplicate code. The reason I've created another contrib module is because it requires ipywidgets and that in turn requires a whole host of other packages (traitlets, ipython.. etc). So if we merge both when a user has only matplotlib and wants to use the normal python file said user will have to install all ipywidgets requirements.

So it might make sense to have 2 modules 1) interactive_guess (current) 2) jupyter where with 1) you only need matplotlib and has the common API and 2) needs ipywidgets and also 1)?

pckroon:
> So it might make sense to have 2 modules 1) interactive_guess (current) 2) jupyter where with 1) you only need matplotlib and has the common API and 2) needs ipywidgets and also 1)?

Exactly what I had in mind. Internally this will mean 1) will get split into two parts, one part dealing with the sliders, and one part dealing exclusively with preparing and updating the matplotlib graphs. It'll require some thought where the split in responsibilities will come exactly: which part decides how many sliders should be made? Their limits? Their labels?

Jhsmit:
It looks like i'm probably not going to have much time to spend on this in the coming 3 months. 

Should the number of sliders not always be equal to the number of (non fixed) parameters?
Then the update function should in both cases be a function that accepts parameters values as kwargs, put evalutates the model, then updates the lines?

What type of visualization do you use for the N dimensional case?

pckroon:
I'm also super swamped until my thesis is finished. 
The number of sliders should (probably) always be the number of parameters. But in N-dimensional cases you can come up with a visualisation strategy where the variables you project out become sliders.

See https://symfit.readthedocs.io/en/stable/examples/ex_interactive_guesses_nD.html for the visualization I use now.

tBuLi:
With regards to visualizing higher dimensional data, perhaps we should look into t-SNE. It seems this is popular in machine learning for visualizing higher dimensional data so it could be what we need: https://scikit-learn.org/stable/auto_examples/manifold/plot_t_sne_perplexity.html#sphx-glr-auto-examples-manifold-plot-t-sne-perplexity-py

Jhsmit:
looks interesting. I found some cool animations: https://distill.pub/2016/misread-tsne/

lesshaste:
I understand https://umap-learn.readthedocs.io/en/latest/ is regarded as much better than tsne these days.

tBuLi:
@lesshaste, that also looks very interesting, thanks for the suggestion! I do think one of these techniques could solve the problem for doing higher dimensional interactive guesses, by playing a game of making the point cloud of the initial guess overlap as much as possible with that of the data before fitting. I don't know it that is guaranteed to be a good initial guess, or that there is still no uniqueness in the output of these algorithms?

PR #268, opened 2019-08-13
Convert unittests to pytests

DuckNrOne:
Hey guys,

sorry for the long time to convert the tests, but now you can find in this repo the new pytests. The converted tests are marked with a *_py* at the end.  I saved the logs of each test just for a better overview. Before you merge it, I would rename tests and delete the logs.

Executing the test, I found many failures at the "old" unittests and that's why not every test executes 100% correctly. Maybe I'll find one or the other reason in the next weeks. Then I would make a new pull request.

Grüße


DuckNrOne:
A little hint to the next commit. Pytest has a little [bug](https://github.com/pytest-dev/pytest/issues/2003). They though the problem is solved, but it isn't apparently.

I solved the problem temporary, but I'm in contact with them.

DuckNrOne:
The last commit fixes #269

tests/test_argument.py
```diff
@@ -23,101 +23,70 @@
...
+    a = Parameter()
+    b = Parameter('b', 10)
+    c = Parameter(value=10)
+    x = Variable()
+    y = Variable('y')
+
+    assert str(a) == '{}_{}'.format(a._argument_name, a._argument_index)
+    assert str(a) == 'par_{}'.format(a._argument_index)
+    assert str(b) != '{}_{}'.format(b._argument_name, b._argument_index)
+    assert str(c) == '{}_{}'.format(c._argument_name, c._argument_index)
```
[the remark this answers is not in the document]
DuckNrOne: I think it's possible to do such little improvements in this file.

symfit/core/fit_results.py
```diff
@@ -65,7 +66,8 @@ def __str__(self):
             value_str = '{:e}'.format(value) if value is not None else 'None'
             stdev = self.stdev(p)
             stdev_str = '{:e}'.format(stdev) if stdev is not None else 'None'
-            res += '{:10}{} {}\n'.format(p.name, value_str, stdev_str, width=20)
+            res += '{:10}{} {}\n'.format(p.name,
+                                         value_str, stdev_str, width=20)
```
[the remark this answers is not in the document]
DuckNrOne: I don't know what you mean with "undo those". I think excluding *fit_result* is the best way to go. Maybe I will look into the file after merging.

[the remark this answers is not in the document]
pckroon: You can try: `git checkout -- fit_results.py`, followed by `git add fit_results.py` and `git commit -m "Undo changes to fit_results.py"`. I *think* that should "reset" the file while preserving your git commits/history for this branch.

tests/test_general.py
```diff
@@ -27,899 +26,941 @@
...
+    """
+    a, b = parameters('a, b')
+    x, y = variables('x, y')
+    new = a*x**2 + b*y**2
+    model = Model(new)
+    ans = model(3, 3, 2, 2)
+    assert isinstance(ans, tuple)
+    z, = ans
+
+    assert z == 36
```
pckroon: I would replace the 36 with `2*3**2 + 2*3**2`. But it can be left for later.
Also note that this test is not that good, since x and y could be swapped, and we'd never see it.

tests/test_general.py
```diff
@@ -27,899 +26,941 @@
...
+
+
+def test_2_gaussian_2d_fitting():
+    """
+    Tests fitting to a scalar gaussian with 2 independent variables with
+    tight bounds.
+    """
+    mean = (0.3, 0.4)  # x, y mean 0.6, 0.4
+    cov = [[0.01**2, 0], [0, 0.01**2]]
+    data = np.random.multivariate_normal(mean, cov, 3000000)
```
pckroon: @tBuLi: Do we really really need 3 million (!) points? Same for the 10k in the test above.

tests/test_general.py
```diff
@@ -27,899 +26,941 @@
...
+    loglikelihood = np.sum(np.log(pdf_i))
+
+    assert fit_result.value(b) / mean == pytest.approx(1, 1e-3)
+    assert fit_result.value(b) / mean == pytest.approx(1, 1e-3)
+    assert fit_result.value(b) / mean == pytest.approx(1, 1e-3)
+    assert fit_result.value(b) / mean == pytest.approx(1, 1e-3)
+    assert fit_result.value(b) / mean == pytest.approx(1, 1e-3)
+    assert fit_result.value(b) / mean == pytest.approx(1, 1e-3)
+    assert fit_result.value(b) / stdev == pytest.approx(1, 1e-3)
+    assert fit_result.stdev(b) / mean_stdev == pytest.approx(1, 1e-3)
```
pckroon: `assert fit_result.stdev(b) == pytest.approx(mean_stdev)`

tests/test_general.py
```diff
@@ -27,899 +26,941 @@
...
+    mean = np.mean(xdata)
+    stdev = np.std(xdata)
+    mean_stdev = stdev/np.sqrt(len(xdata))
+
+    fit = Fit(pdf, xdata, objective=LogLikelihood)
+    fit_result = fit.execute()
+
+    assert fit_result.value(mu) / mean == pytest.approx(1, 1e-6)
+    assert fit_result.stdev(mu) / mean_stdev == pytest.approx(1, 1e-3)
+    assert fit_result.value(sig) / np.std(xdata) == pytest.approx(1, 1e-6)
```
pckroon: `assert value == approx(reference)`

tests/test_general.py
```diff
@@ -27,899 +26,941 @@
...
+    fit = Fit(pdf, x=xdata, y=ydata, objective=LogLikelihood)
+    fit_result = fit.execute()
+
+    assert fit_result.value(x0) / mean[0] == pytest.approx(1, 1e-2)
+    assert fit_result.value(y0) / mean[1] == pytest.approx(1, 1e-2)
+    assert fit_result.value(
+        sig_x) / np.sqrt(cov[0, 0]) == pytest.approx(1, 1e-2)
+    assert fit_result.value(
+        sig_y) / np.sqrt(cov[1, 1]) == pytest.approx(1, 1e-2)
+    assert fit_result.value(rho) / r == pytest.approx(1, 1e-2)
```
pckroon: `assert value == approx(reference)`

tests/test_general.py
```diff
@@ -27,899 +26,941 @@
...
+    # h_smooth = np.linspace(0,60,100)
+    # t_smooth = t_model(y=h_smooth, **fit_result.params)
+
+    # Lets with the results from curve_fit, no weights
+    popt_noweights, pcov_noweights = curve_fit(
+        lambda y, p: (2 * y / p)**0.5, y_data, t_data)
+
+    assert fit_result.value(g) == pytest.approx(popt_noweights[0])
+    assert fit_result.stdev(g) == pytest.approx(
+        np.sqrt(pcov_noweights[0, 0]), 1e-6)
```
pckroon: If you can do `approx(numpy_array)` it might be an idea to replace all `numpy.testing.assert_almost_equal` calls. Depending on your mood/time we can also leave it for later.

tests/test_general.py
```diff
@@ -27,899 +26,941 @@
...
+    fit = Fit(model, x=xdata, y=ydata, z=zdata, absolute_sigma=False)
+    fit_result = fit.execute()
+
+    # Same as Mathematica default behavior.
+    assert fit_result.value(a) / 2.9956 == pytest.approx(1, 1e-4)
+    assert fit_result.value(b) / 0.563212 == pytest.approx(1, 1e-4)
+    assert fit_result.value(c) / 3.59732 == pytest.approx(1, 1e-4)
+    assert fit_result.stdev(a) / 0.278304 == pytest.approx(1, 1e-4)
+    assert fit_result.stdev(b) / 0.224107 == pytest.approx(1, 1e-4)
+    assert fit_result.stdev(c) / 0.980352 == pytest.approx(1, 1e-4)
```
pckroon: `assert value == approx(reference)`

tests/test_general.py
```diff
@@ -27,899 +26,941 @@
...
+    fit = Fit(model, xdata, ydata, zdata, absolute_sigma=True)
+    fit_result = fit.execute()
+    # Same as Mathematica in Measurement error mode, but without suplying
+    # any errors.
+    assert fit_result.value(a) / 2.9956 == pytest.approx(1, 1e-4)
+    assert fit_result.value(b) / 0.563212 == pytest.approx(1, 1e-4)
+    assert fit_result.value(c) / 3.59732 == pytest.approx(1, 1e-4)
+    assert fit_result.stdev(a) / 0.643259 == pytest.approx(1, 1e-4)
+    assert fit_result.stdev(b) / 0.517992 == pytest.approx(1, 1e-4)
+    assert fit_result.stdev(c) / 2.26594 == pytest.approx(1, 1e-4)
```
pckroon: `assert value == approx(reference)`

tests/test_general.py
```diff
@@ -27,899 +26,941 @@
...
+    x, y_1, y_2 = variables('x, y_1, y_2')
+    a, b = parameters('a, b')
+    # This way the test fails rather than errors.
+    try:
+        Model({
+            y_1: 2 * a * x,
+            y_2: b * x**2
+        })
+    except Exception as error:
+        self.fail('test_model_from_dict raised {}'.format(error))
```
pckroon: `self.fail` will not work. Should probably be `pytest.fail`

tests/test_minimize.py
```diff
@@ -15,252 +15,254 @@
...
+    res = minimize(func, [-1.0, 1.0], args=(-1.0,), jac=func_deriv,
+                   method='BFGS', options={'disp': False})
+    fit = Fit(model=- model)
+    assert isinstance(fit.objective, MinimizeModel)
+    assert isinstance(fit.minimizer, BFGS)
+
+    fit_result = fit.execute()
+
+    assert fit_result.value(x) / res.x[0] == pytest.approx(1.0, 1e-6)
+    assert fit_result.value(y) / res.x[1] == pytest.approx(1.0, 1e-6)
```
pckroon: `assert value == approx(reference)`

tests/test_minimize.py
```diff
@@ -15,252 +15,254 @@
...
+
+    # Same test, but with constraints in place.
+    res = minimize(func, [-1.0, 1.0], args=(-1.0,), jac=func_deriv,
+                   constraints=cons, method='SLSQP', options={'disp': False})
+
+    from symfit.core.minimizers import SLSQP
```
pckroon: Import is not used?

tests/test_minimize.py
```diff
@@ -15,252 +15,254 @@
...
+    fit = Fit(- model, constraints=constraints)
+    assert fit.constraints[0].constraint_type == Ge
+    assert fit.constraints[1].constraint_type == Eq
+    fit_result = fit.execute()
+    assert fit_result.value(x), res.x[0] == pytest.approx(1e-6)
+    assert fit_result.value(y), res.x[1] == pytest.approx(1e-6)
```
pckroon: This got typoed somewhere. Should probably be:
`assert fit_result.value(y) == approx(res.x[1], 1e-6)`

tests/test_minimize.py
```diff
@@ -15,252 +15,254 @@
...
-        x1, x2 = parameters('x1, x2', value=x0)
-        with self.assertRaises(TypeError):
-            fit = BasinHopping(
-                func2d_symfit, [x1, x2],
-                local_minimizer=NelderMead(func2d_symfit, [x1, x2],
-                                           jacobian=jac2d_symfit)
-            )
+            assert False
+        except ModelError:
+            assert True
```
pckroon: Should maybe be:
```python3
try:
    model = ...
except ModelError:
    pytest.success()
else:
    pytest.fail()
```
Or maybe even better is assure it raises:
```python3
with pytest.assert_raises(ModelError):
    model = ...
```

tests/test_minimize.py
```diff
@@ -15,252 +15,254 @@
...
+        func2d, x0, minimizer_kwargs=minimizer_kwargs, niter=200)
+
+    np.random.seed(555)
+    x1, x2 = parameters('x1, x2', value=x0)
+    try:
         fit = BasinHopping(
             func2d_symfit, [x1, x2],
-            local_minimizer=BFGS(func2d_symfit, [x1, x2], jacobian=jac2d_symfit)
+            local_minimizer=NelderMead(func2d_symfit, [x1, x2],
+                                       jacobian=jac2d_symfit)
```
pckroon: This is a good linebreak by the way

tests/test_minimize.py
```diff
@@ -15,252 +15,254 @@
...
-        unittest.main(warnings='ignore')
-        # Note that unittest will catch and handle exceptions raised by tests.
-        # So this line will *only* deal with exceptions raised by the line
-        # above.
+        assert False
     except TypeError:
-        # In Py2, unittest.main doesn't take a warnings argument
-        warnings.simplefilter('ignore')
-        unittest.main()
+        assert True
```
pckroon: Should probably be `with pytest.raises(TypeError)`

tests/test_minimize.py
```diff
@@ -15,252 +15,254 @@
...
+    fit = BasinHopping(
+        func2d_symfit, [x1, x2],
+        local_minimizer=BFGS(func2d_symfit, [x1, x2], jacobian=jac2d_symfit)
+    )
+    fit_result = fit.execute(niter=200)
+    assert isinstance(fit.local_minimizer.jacobian, MinimizeModel)
+    assert isinstance(fit.local_minimizer.jacobian.model,
+                      CallableNumericalModel)
+    assert res.x[0] / fit_result.value(x1) == 1.0
+    assert res.x[1] / fit_result.value(x2) == 1.0
```
pckroon: `assert value == reference`

tests/test_minimizers.py
```diff
@@ -53,7 +59,8 @@ def eval_hessian(self, *args, **kwargs):
         chi2_jac = super(SqrtLeastSquares, self).eval_jacobian(*args, **kwargs)
         chi2_hess = super(SqrtLeastSquares, self).eval_hessian(*args, **kwargs)
         return - 0.5 * (1 / chi2) * np.outer(sqrt_chi2_jac, chi2_jac) \
-               + 0.5 * (1 / sqrt_chi2) * chi2_hess
+            + 0.5 * (1 / sqrt_chi2) * chi2_hess
```
pckroon: Should probably be just on 1 line.

tests/test_minimizers.py
```diff
@@ -73,346 +80,334 @@ def subclasses(base, leaves_only=True):
...
+
+    def f(x, a, b):
+        return a * x + b
+
+    def chi_squared(a, b):
+        return np.sum((ydata - f(xdata, a, b))**2)
+
+    with warnings.catch_warnings(record=True) as w:
+        # Should no longer raise warnings, because internally we practice
+        # what we preach.
```
pckroon: I don't understand why/what this does. Maybe @tBuLi remembers? But not for this PR.

tests/test_minimizers.py
```diff
@@ -73,346 +80,334 @@ def subclasses(base, leaves_only=True):
...
+        fit_custom = BFGS(chi_squared, [a, b])
+        assert len(w) == 0
+
+    fit_custom_result = fit_custom.execute()
+
+    assert isinstance(fit_custom_result, FitResults)
+    assert fit_custom_result.value(
+        a) / fit_result.value(a) == pytest.approx(1.0, 1e-5)
+    assert fit_custom_result.value(
+        b) / fit_result.value(b) == pytest.approx(1.0, 1e-4)
```
pckroon: Linebreaks, and `assert value == approx(reference)`

tests/test_minimizers.py
```diff
@@ -73,346 +80,334 @@ def subclasses(base, leaves_only=True):
...
+    assert callable_model.interdependent_vars == []
+    assert callable_model.connectivity_mapping == {y: {a, b}}
+    fit_custom = BFGS(callable_model, [a, b])
+    fit_custom_result = fit_custom.execute()
+
+    assert isinstance(fit_custom_result, FitResults)
+    assert fit_custom_result.value(
+        a) / fit_result.value(a) == pytest.approx(1.0, 1e-5)
+    assert fit_custom_result.value(
+        b) / fit_result.value(b) == pytest.approx(1.0, 1e-4)
```
pckroon: Linebreaks and `assert value == approx(reference)`

tests/test_minimizers.py
```diff
@@ -73,346 +80,334 @@ def subclasses(base, leaves_only=True):
...
+                    # These attr are new instances, and therefore do not
+                    # pass an equality test. All we can do is see if they
+                    # are at least the same type.
+                    if isinstance(value, (list, tuple)):
+                        for val1, val2 in zip(value, new_value):
+                            assert isinstance(val1, val2.__class__)
+                            if key == 'constraints':
+                                assert val1.model.constraint_type == val2.model.constraint_type
+                                assert list(val1.model.model_dict.values())[
+                                    0] == list(val2.model.model_dict.values())[0]
```
pckroon: linebreak

tests/test_minimizers.py
```diff
@@ -73,346 +80,334 @@ def subclasses(base, leaves_only=True):
...
+                                if isinstance(val1, dict):
+                                    assert val1['type'] == val2['type']
+                                    assert set(val1.keys()) == set(val2.keys())
+                                elif isinstance(val1, NonlinearConstraint):
+                                    # For trust-ncg we manually check if
+                                    # their dicts are equal, because no
+                                    # __eq__ is implemented on
+                                    # NonLinearConstraint
+                                    assert len(val1.__dict__) == len(
+                                        val2.__dict__)
```
pckroon: linebreak

tests/test_minimizers.py
```diff
@@ -73,346 +80,334 @@ def subclasses(base, leaves_only=True):
...
+                                    for key in val1.__dict__:
+                                        try:
+                                            assert val1.__dict__[
+                                                key] == val2.__dict__[key]
```
pckroon: linebreak

tests/test_minimizers.py
```diff
@@ -73,346 +80,334 @@ def subclasses(base, leaves_only=True):
...
+                    raise err
+        assert set(fit.__dict__.keys()) == set(pickled_fit.__dict__.keys())
+
+        # Test if we converge to the same result.
+        np.random.seed(2)
+        res_before = fit.execute()
+        np.random.seed(2)
+        res_after = pickled_fit.execute()
+        assert FitResults._array_safe_dict_eq(
+            res_before.__dict__, res_after.__dict__)
```
pckroon: linebreak

tests/test_model.py
```diff
@@ -18,435 +18,461 @@
...
+    model = Model(model_dict)
 
-    def test_neg(self):
-        """
-        Test negation of all model types
-        """
-        x, y_1, y_2 = variables('x, y_1, y_2')
-        a, b = parameters('a, b')
+    assert id(model[y_1]) == id(model_dict[y_1])
+    assert id(model[y_2]) == id(model_dict[y_2])
```
pckroon: `model[y_1] is model_dict[y_1]`. But can be left for later

tests/test_model.py
```diff
@@ -18,435 +18,461 @@
...
+    model = Model(model_dict)
+
+    model_neg = - model
+    for key in model:
+        if key in model.dependent_vars:
+            assert model[key] == - model_neg[key]
+        elif key in model.interdependent_vars:
+            assert model[key] == model_neg[key]
+        else:
+            raise Exception('There should be no such variable')
```
pckroon: pytest.fail would be better, since then the test F(ails) instead of E(rrors)

tests/test_model.py
```diff
@@ -18,435 +18,461 @@
...
+                                          [], [a, b])
+    assert not model.__signature__ == faulty_model.__signature__
+
+    try:
+        # This is an incorrect signature, even though the lambda function is
+        # correct. Should fail.
+        faulty_model(xdata, 5.5, 15.0)
+        assert False
+    except TypeError:
+        assert True
```
pckroon: `pytest.raises`

tests/test_model.py
```diff
@@ -18,435 +18,461 @@
...
+    faulty_model = CallableNumericalModel(
+        {y: lambda x, a, b: a * x + b, z: lambda x, a: x**a}, [x], [a, b]
+    )
+    assert model.__signature__ == faulty_model.__signature__
+    try:
+        # Lambda got an unexpected keyword 'b'
+        faulty_model(xdata, 5.5, 15.0)
+        assert False
+    except TypeError:
+        assert True
```
pckroon: `pytest.raises`

tests/test_model.py
```diff
@@ -18,435 +18,461 @@
...
+    faulty_model = CallableNumericalModel(
+        {y: lambda x, a, c=5: a * x + c}, [x], [a, b]
+    )
+    assert model.__signature__ == faulty_model.__signature__
+    try:
+        # Lambda got an unexpected keyword 'b'
+        faulty_model(xdata, 5.5, 15.0)
+        assert False
+    except TypeError:
+        assert True
```
pckroon: `pytest.raises`

tests/test_model.py
```diff
@@ -18,435 +18,461 @@
...
-        for var, func in callable_model.vars_as_functions.items():
-            self.assertEqual(
-                set(str(x) for x in callable_model.connectivity_mapping[var]),
-                set(str(x.__class__) if isinstance(x, Function) else str(x)
-                    for x in func.args)
-            )
+        model = Model(model_dict)
+        assert False
+    except ModelError:
+        assert True
```
pckroon: `pytest.raises`

tests/test_model.py
```diff
@@ -18,435 +18,461 @@
...
-        )
-        self.assertEqual(
-            jac_model.model_dict,
+    try:
+        # Invalid, parameters can not be keys
+        model_dict = {c: a ** 3 * x + b ** 2}
+        model = Model(model_dict)
+        assert False
+    except ModelError:
+        assert True
```
pckroon: `pytest.raises`

tests/test_objectives.py
```diff
@@ -23,268 +23,259 @@
...
+            incomplete_data = {x: xdata, y: ydata}
+            data = {x: xdata, y: ydata,
+                    model.sigmas[y]: np.ones_like(ydata)}
+            overcomplete_data = {x: xdata, y: ydata, z: ydata,
+                                 model.sigmas[y]: np.ones_like(ydata)}
+        try:
+            obj = objective(model, data=incomplete_data)
+            assert False
+        except KeyError:
+            assert True
```
pckroon: `pytest.raises`

tests/test_ode.py
```diff
@@ -10,210 +10,225 @@
...
+    ode_result = fit.execute()
+
+    c1.value = ydata[0]
+    fit = Fit(sol, t=tdata, y=ydata)
+    fit_result = fit.execute()
+
+    assert ode_result.value(p) / fit_result.value(p) == pytest.approx(1, 1e-2)
+    assert ode_result.r_squared / \
+        fit_result.r_squared == pytest.approx(1, 1e-4)
+    assert ode_result.stdev(p) / fit_result.stdev(p) == pytest.approx(1, 1e-2)
```
pckroon: `assert value == approx(reference)`

tests/test_ode.py
```diff
@@ -10,210 +10,225 @@
...
+        fit_result.r_squared == pytest.approx(1, 1e-4)
+
+    fit = Fit(ode_model, t=tdata, a=adata, b=None)
+    ode_result = fit.execute()
+    assert ode_result.value(
+        k) / fit_result.value(k) == pytest.approx(1.0, 1e-4)
+    assert ode_result.stdev(
+        k) / fit_result.stdev(k) == pytest.approx(1.0, 1e-4)
+    assert ode_result.r_squared / \
+        fit_result.r_squared == pytest.approx(1, 1e-4)
```
pckroon: Linebreaks and `assert value == approx(reference)`

tests/test_ode.py
```diff
@@ -10,210 +10,225 @@
...
+    fit_result = fit.execute()
+
+    fit = Fit(ode_model, t=tdata, a=adata, b=None, minimizer=MINPACK)
+    ode_result = fit.execute()
+    assert ode_result.value(
+        k) / fit_result.value(k) == pytest.approx(1.0, 1e-4)
+    assert ode_result.stdev(
+        k) / fit_result.stdev(k) == pytest.approx(1.0, 1e-4)
+    assert ode_result.r_squared / \
+        fit_result.r_squared == pytest.approx(1, 1e-4)
```
pckroon: Linebreaks and `assert value == approx(reference)`

tests/test_ode.py
```diff
@@ -10,210 +10,225 @@
...
+    k, a0 = parameters('k, a0')
+
+    model_dict = {
+        D(a, t): - k * a * t,
+    }
+    try:
+        fit = Fit(model_dict, t=tdata, a=adata)
+        assert False
+    except RuntimeWarning:
+        assert True
```
pckroon: `pytest.raises`

tests/test_support.py
```diff
@@ -19,219 +19,215 @@
...
+    assert x2.value == 1.3
+    assert x1.min == 0.0
+    assert x2.min == 0.0
+    assert not x1.fixed
+    assert not x2.fixed
+    with pytest.raises(ValueError):
+        x1, x2 = parameters('x1, x2', value=[2.0, 1.3, 3.0], min=0.0)
+
+    x1, x2 = parameters('x1, x2', value=[
+                        2.0, 1.3], min=[-30, -10], max=[300, 100], fixed=[True, False])
```
pckroon: Linebreak. You can break between the keyword arguments if you want. E.g.
```
parameters('x1, 
           value=[2, 1.3],
           min=[-30, -10],
           ...
           )
```
or
```
parameters(
    'x1, x2',
    value=[2, 1.3],
    min=[-30, -10],
    ...
)
```

symfit/core/fit_results.py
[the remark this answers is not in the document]
DuckNrOne: So you want to have the old unittest-type for this file, right? :D

DuckNrOne:
Everything looks fine! I'm happy about such constructive feedback :D! Nobody learns to finish, especially in this field ;)

In one or two weeks I have time to look into the code searching for the lines. if I ask, I know who to contact.

symfit/core/fit_results.py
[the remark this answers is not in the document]
pckroon: There's no tests in that file, so it shouldn't be part of the PR ;)

tests/test_argument.py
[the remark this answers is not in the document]
tBuLi: For this one I'm not sure you could do that actually, since if the number of tests is changed the number of the par_x is no longer the same :(.

tests/test_general.py
tBuLi: Yeah if I remember correctly we did, multivariate Gaussian fitting is a bitch. We could off course cherry pick and find a random seed for which less also works... :P

tests/test_minimizers.py
tBuLi: Not sure, but I think it is because we changed the API for directly using the minimizers to raise a deprecation warning when used old school. This is just to make sure our test is correct. Should of course be changed to a test for the deprecation warning instead...

tests/test_general.py
pckroon: Hmmmn. Definitely out-of-scope for this PR, but wouldn't it be equivalent to evaluating 100 points on a gaussian curve since we bin the points later on? You can then put some uncertainty on those values, but that requires more maths than I'm currently capable of.

tests/test_argument.py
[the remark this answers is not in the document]
DuckNrOne: You mean I should skip this, right? @tBuLi

tests/test_general.py
```diff
@@ -27,899 +26,941 @@
...
+        a_i=xdata[0],
+        b_i=xdata[1],
+        c_i=xdata[2],
+        minimizer=MINPACK
+    )
+    fit_result = fit.execute()
+
+    assert fit_result.value(b) / 1.006143e+02 == pytest.approx(1.0, 1e-4)
+    assert fit_result.value(c) / 7.085713e+01 == pytest.approx(1.0, 1e-5)
+    assert fit_result.value(a) / 9.985691 == pytest.approx(1.0, 1e-5)
```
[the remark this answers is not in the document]
DuckNrOne: I can replace the other numbers with `np.mean(xdata[])` too

tests/test_general.py
DuckNrOne: Allright. I'll skip that first

tests/test_general.py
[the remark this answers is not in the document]
pckroon: Sounds great!

tests/test_minimize.py
DuckNrOne: It's already imported in the header

tests/test_model.py
DuckNrOne: It's not a big thing, right? I changed it :D

pckroon: Awesome :)

tests/test_general.py
DuckNrOne: I'll replace every `numpy.testing.assert_almost_equal` and `numpy.testing.assert_allclose` if two arguments are used (not three or more).

pckroon: How about `assert np.allclose(arr1, arr2)`?

DuckNrOne: Pytest cannot compare in the test *model* actual models, but numpy. So I let it stand at two lines

tests/test_argument.py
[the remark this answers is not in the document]
DuckNrOne: The I think I have to skip this one :)

tests/test_general.py
DuckNrOne: numpy can comare lists, pytest not

tests/test_minimizers.py
DuckNrOne: So how can I handle this one? :D

tests/test_model.py
```diff
@@ -110,46 +110,40 @@ def test_CallableNumericalModel():
 
     xdata = np.linspace(0, 10)
     ydata = model(x=xdata, a=5.5, b=15.0).y + np.random.normal(0, 1)
+
+    # TODO #assert model(x=xdata, a=5.5, b=15.0) == pytest.approx(numerical_model(x=xdata, a=5.5, b=15.0))
     np.testing.assert_almost_equal(
         model(x=xdata, a=5.5, b=15.0),
-        numerical_model(x=xdata, a=5.5, b=15.0),
+        numerical_model(x=xdata, a=5.5, b=15.0)
     )
```
pckroon: I like the TODO in case this doesn't work, but could you still try `assert np.array(model(...)) == approx(np.array(numerical_model(...))`

pckroon:
Nice work so far :)
However, you've reverted your changes to test_fit_result.py which is probably undesirable. You /do/ want to reset your changes to fit_result.py.

DuckNrOne:
> Nice work so far :)
> However, you've reverted your changes to test_fit_result.py which is probably undesirable. You /do/ want to reset your changes to fit_result.py.

I thought I should skip the whole test?

tests/test_minimizers.py
pckroon: Ignore it for now.

pckroon:
>> Nice work so far :)
>> However, you've reverted your changes to test_fit_result.py which is probably undesirable. You /do/ want to reset your changes to fit_result.py.

> I thought I should skip the whole test?

No, changes to `tests/test_fit_results.py` are desirable. Your PR also made changes to `symfit/core/fit_results.py`, which *are* unrelated to this PR.

DuckNrOne:
> > > Nice work so far :)
> > > However, you've reverted your changes to test_fit_result.py which is probably undesirable. You /do/ want to reset your changes to fit_result.py.
> 
> > I thought I should skip the whole test?
> 
> No, changes to `tests/test_fit_results.py` are desirable. Your PR also made changes to `symfit/core/fit_results.py`, which _are_ unrelated to this PR.

Oh 😄 I mistook the folder. I'll change this

DuckNrOne:
Thank you for your OK :)
But before you merge, I would like to implement`fixtures` and `parametrizes` to use pytest the right way

pckroon:
No. Do that in a next PR to keep them small/contained/managable/reviewable. And in a next PR, let's do one file at the time for the same reasons.

DuckNrOne:
Allright. I could open PR for every File tm, okay?

pckroon:
That sounds good to me :)
You should be able to create a branch of of this one so you don't have to wait for us to merge this (but I'll poke @tBuLi a bit). It could be that git will get confused once we rebase this PR to clean up the commit history a bit though.

DuckNrOne:
I would create some branches form master and add the tests peace by peace. It's the easiest way, I think :)

pckroon:
That would mean you won't have the changes you made in this PR which at the very least causes a duplication of effort, and in the worst case a very very difficult merge.

DuckNrOne:
If I open a PR for single tests? really? :)

pckroon:
That would make the PRs really very small. It would be preferably to large PRs though... Maybe a few tests per PR, and at most a file?

DuckNrOne:
One big PR with every test as a commit? Would make more sense in my opinion :)

pckroon:
I'd still prefer to limit it to one PR per file. But I can review it commit by commit.

DuckNrOne:
If I understand you the right way: One branch and I commit test by test and after every commit, I open a PR? No right? :D Can you make some introductions please :))

tBuLi:
Thank for your good work! I think @pckroon means that in order to keep the chunks manageable for review it is better to divide it up into several PRs. I agree that one PR per file is probably a good idea.

I'll try to go over this in the coming days, but I know that @pckroon has already reviewed this extensively and is more experienced with pytest than I am so if he's happy I probably wont have a lot of critique to add :P.

DuckNrOne:
I could open a PR for every test and  @tBuLi could review (and possibly merdge or critique) them peace by peace.

pckroon:
Something like that would work for me (except that I'd also be happy to review). I'm OK with PRs going beyond 4 tests if you think it makes sense, so long as they stay under ~1 file.
Also, as far as I'm concerned you're free to make fairly radical changes. I can foresee you having to merge or split tests, as well as deciding that a specific test should be replaced by another so that it tests the same behaviour, but now fits in the same `pytest.mark.parametrize` shape as another. Just make sure the coverage doesn't drop.

Happy coding :)

DuckNrOne:
I would recomment to merge these pure converted tests in multiple PRs (2-4 tests per PR). Later on I improve some of them with fixtures and parametrizes and create new PRs.
In the sense of convert first and improve second...two steps

DuckNrOne:
Or I create PRs for all tests except `test_argument`, `test_fit_result`, `test_global_opt`, `test_auto_fit` and `test_constrained`. I would continue to work on the tests and add fixtures and parametrizes.

tBuLi:
Just to let you know, I plan to go through this PR on Friday in more detail, and hopefully merge it if there are no big things left.

tests/test_argument.py
```diff
@@ -23,101 +23,70 @@
...
+def test_slots():
+    """
+    Make sure Parameters and Variables don't have a __dict__
+    """
+    P = Parameter('P')
+
+    # If you only have __slots__ you can't set arbitrary attributes, but
+    # you *should* be able to set those that are in your __slots__
     try:
-        unittest.main(warnings='ignore')
```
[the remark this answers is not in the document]
pckroon: Pytest has no equivalent to unittest.main. There is more fine-grained control of what test should display/ignore what warning, but I'd say it's out of scope for now.

https://docs.pytest.org/en/latest/warnings.html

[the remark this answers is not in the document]
pckroon: Alternatively, we can tell pytest to ignore *all* warnings. Although that's probably also now what we want.

DuckNrOne:
I'm at the moment on a 6 day bike tour...I'm afraid that I can not work on it until Sunday.

tests/test_argument.py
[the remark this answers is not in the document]
DuckNrOne: If I read the docs right, it's possible to ignore warnings for single methods with `@pytest.mark.filterwarnings("ignore:XXX")`. That would be a possibility.

tests/test_argument.py
```diff
@@ -4,6 +4,7 @@
 import sys
 import sympy
 import warnings
+warnings.filterwarnings("ignore")
```
tBuLi: why is this needed?

tBuLi: Ah, never mind I just fould your comment about this above :)

pckroon:
I know I'm starting to be annoying, but I think I'd be happier if we just let the warning be and fill the output. They're there for a reason after all. Later, once we've been over the tests in a few PRs we can always choose which, if not all, we want to squelch.

tBuLi:
I agree with @pckroon on that, I would also not ignore warnings yet for now.

And something nitpicky I'm noticing now upon going through it in detail is that there are many lines which pass 80 characters with just a few characters, but that are broken into two since it violates PEP 8. But I disagree with that: as long as a line is just a couple of characters over then it is better to leave it as one line, since one line represents one coherent block of information. Splitting it in two actually makes it harder to read, which is against the intention of python for me.

For example:
```python
assert fit_result.value(
        x0) / np.mean(data[:, 0]) == pytest.approx(1.0, 1e-2)
```
is much harder to read than 
```python
assert fit_result.value(x0) / np.mean(data[:, 0]) == pytest.approx(1.0, 1e-2)
```
So I would be very happy if those kinds of lines are restored to a single line. As I'm now going over it, I'll make a PR against your branch so you don't have to redo the work.

pckroon:
> And something nitpicky I'm noticing now upon going through it in detail is that there are many lines which pass 80 characters with just a few characters, but that are broken into two since it violates PEP 8. But I disagree with that: as long as a line is just a couple of characters over then it is better to leave it as one line, since one line represents one coherent block of information. Splitting it in two actually makes it harder to read, which is against the intention of python for me.

Are you sure you're reviewing the correct code? Most if not all of those were fixed in later commits.

DuckNrOne:
Allright. So I'll bring back the warnings and look over the linebreaks in the next few days

DuckNrOne:
Allright. So I'll bring back the warnings today.

tBuLi:
> > And something nitpicky I'm noticing now upon going through it in detail is that there are many lines which pass 80 characters with just a few characters, but that are broken into two since it violates PEP 8. But I disagree with that: as long as a line is just a couple of characters over then it is better to leave it as one line, since one line represents one coherent block of information. Splitting it in two actually makes it harder to read, which is against the intention of python for me.
> 
> Are you sure you're reviewing the correct code? Most if not all of those were fixed in later commits.

I just checked out this branch this morning, didn't have it local before.

> Allright. So I'll bring back the warnings and look over the linebreaks in the next few days

Don't worry about the line breaks, I'll make a PR against your branch with anything I catch right now.

tests/test_model.py
```diff
@@ -18,435 +18,439 @@
...
+    fit = Fit(numerical_model, x=xdata, y=ydata, z=zdata)
+    numerical_result = fit.execute()
+    for param in [a, b]:
+        assert mixed_result.value(param) == pytest.approx(
+            numerical_result.value(param))
+
+        if mixed_result.stdev(param) is  None and numerical_result.stdev(param) is None:
+            assert True
+        elif mixed_result.stdev(param) or  None and numerical_result.stdev(param) is None:
+            assert False
```
tBuLi: I assume this is needed because pytes.approx t can't deal with None. If so, this can be replaced by 
```python
if mixed_result.stdev(param) is None or numerical_result.stdev(param) is None:
    assert mixed_result.stdev(param) == numerical_result.stdev(param)
```

tests/test_model.py
```diff
@@ -18,435 +18,439 @@
...
+        assert False
+    else:
+        assert fit_result.stdev(a) == pytest.approx(flat_result.stdev(a))
+    
+    if fit_result.stdev(b) is None and flat_result.stdev(b) is None:
+        assert True
+    elif fit_result.stdev(b) is None or flat_result.stdev(b) is None:
+        assert False
+    else:
+        assert fit_result.stdev(b) == pytest.approx(flat_result.stdev(b))
```
tBuLi: The same thing applies here: why is this suddenly split out between None or not None? The result should always be deterministic, if not we have other problems. So only the pytest.approx should be enough here.

tests/test_objectives.py
```diff
@@ -23,268 +23,253 @@
...
+
+    # TODO: These shapes should not have the ones! This is due to the current
+    # convention that scalars should be returned as a 1d array by Model's.
+    assert eval_exact[0].shape == (1,)
+    assert jac_exact[0].shape == (2, 1)
+    assert hess_exact[0].shape == (2, 2, 1)
+    # Test if identical to MinimizeModel
+    assert eval_exact[0] == pytest.approx(eval_minimizemodel)
+    assert jac_exact[0] == pytest.approx(jac_minimizemodel)
+    assert hess_exact[0] == pytest.approx(hess_minimizemodel)
```
tBuLi: Does this do the same thing as the original numpy.testing.assert_almost_equal?

tBuLi:
Thanks for all your work, on the whole I'm quite satisfied and I think we're very close to being able to merge this. I made some comments on your code where I think some discussion might be needed, and I made a PR against yours with some simple changes that I have spotted.

But more importantly, the following files have not been translated in the same way that the other have been:
- test_fit_result.py
- test_global_opt.py
It still contains an object with all the tests on it, rather than having the tests as separate functions. Those really need fixtures like @pckroon suggested. After that has been done, I'm happy!

DuckNrOne:
I would like to improve the tests with fixtures and parametrizes. But I doubt that it'll finish before the weekend.

tBuLi:
That's fine of course, do it at your earliest convenience :).

tests/test_model.py
DuckNrOne: I tried to compare it with `unittest`. If both stdevs are None, there are equal. For `unittest` it's possible to compare it, but pytest doesn't (like you said abowe). In this spesific test there is a case, where both result are `None`. It was in my eyes a good way to compensate for the missing feature of pytest.

tBuLi: Good point. But then I think that in hindsight a better test would be to just check that it is really `None`, since for `CallableNumericalModel` this is what we would expect. If we somehow improve that in the future than we will just just change it then.

So could you change this to an explicit check that they are all `None`?

tests/test_objectives.py
DuckNrOne: `numpy.testing.assert_almost_equal` can compare lists of lists, but when you try to compare with `pytest.approx`, you get an TypeError.

tests/test_model.py
DuckNrOne: So you suggest checking if the `CallableNumericalModel` is `None` right? And if so, to skip the test with this model?

tests/test_model.py
DuckNrOne: I implement the check for `None` because I got an error from the pytest. Apparently, there is a case with `None`. But I see now, that the implementation ist wrong :D I changed an `or` with an `is` and a `and` with an `or`:
```python
if mixed_result.stdev(param) is  None and numerical_result.stdev(param) is None:
    assert True
elif mixed_result.stdev(param) is None or numerical_result.stdev(param) is None:
    assert False
else:
    assert mixed_result.stdev(param) == pytest.approx(numerical_result.stdev(param))
```
I would prefer the following implementation, because if only one object is `None` the test has to fail. In your case, you would try to compare the object, if at least one of them is `None`. 
```python
if mixed_result.stdev(param) is not None and numerical_result.stdev(param) is not None:
    assert mixed_result.stdev(param) == pytest.approx(numerical_result.stdev(param))
else
    assert  mixed_result.stdev(param) is None and numerical_result.stdev(param) is None
```
or would you suggest skipping the test if more than one object is `None`?

tests/test_model.py
tBuLi: No, I mean changing the relevant test to:
```python
assert fit_result.stdev(b) is None and flat_result.stdev(b) is None
```
to actually not allow anything else.

DuckNrOne:
The coverage dopted because we don't check for `NoneType` and `fit_result.stdev(b) == pytest.approx(flat_result.stdev(b))` anymore.

DuckNrOne:
I'm working at the `test_fit_result.py` at the moment.
I noticed that during the test *test_contest_objective_included* `chained_result` is not listed (I added `LeastSquares`). There are such small gaps again and again :D
- *test_pickle*: `minpack_result``is missing
- *test_gof_presence*: `chained_result`, `constrained_result` and `constrained_basinhopping_result` are missing

I was able to fill the gap at`test_contest_objective_included` but I would need your help with the rest :)
By filling the gaps the coverage would increase and the use of `parametrize` and `fixtures` would be cleaner.

DuckNrOne:
Since #274  was merged, I would suggest closing this one :)

PR #279, opened 2019-10-30
Allow Parameters as initial values for ODEModel

### LINKED ISSUE #160: Parameters should be allowed as initial values to ODEModel
In relation to [this SO](https://stackoverflow.com/questions/49149241/ode-fitting-with-symfit-for-python-how-to-get-estimations-for-intial-values) question, it might be worth re-investigating how to allow Parameters to be allowed as initial values, e.g.

```python
x0 = Parameter('x0')
ode_model = ODEModel(model_dict, initial={t: 0.0, x: x0, s: 20.0, p: 0.0})
```
A quick scan of the source tells me this should be straightforward, it will require one extra line to dump any Parameters from initial into the list of parameters of the model, and an extra line or two in `__call__` to correctly feed the provided guesses to the ode-solver.

pckroon:
Fixes #160


symfit/core/models.py
```diff
@@ -1070,6 +1070,9 @@ def eval_components(self, *args, **kwargs):
         Dfun = lambda ys, t, *a: [[c(t, *(list(ys) + list(a))) for c in row] for row in self._njacobian]
 
         initial_dependent = [self.initial[var] for var in self.dependent_vars]
+        for idx, init_var in enumerate(initial_dependent):
```
tBuLi: this needs a comment

symfit/core/models.py
```diff
@@ -1100,7 +1103,7 @@ def eval_components(self, *args, **kwargs):
             initial_dependent,
             t_bigger,
             args=tuple(
-                bound_arguments.arguments[param.name] for param in self.params),
+                bound_arguments.arguments[param.name] for param in self.model_params),
```
tBuLi: why was this switch needed? I'm guessing because the parameter corresponding to the initial value should not be provided here? I feel like this needs either a comment or a better variable name.

tests/test_ode.py
```diff
@@ -97,6 +97,40 @@ def test_polgar(self):
...
+        print(results)
+        self.assertAlmostEqual(a0.value, 10)
+
+        # plt.plot(tdata, AA, color='red', label='[AA]')
+        # plt.plot(tdata, AAB, color='blue', label='[AAB]')
+        # plt.plot(tdata, BAAB, color='green', label='[BAAB]')
+        # plt.plot(tdata, b(d=BAAB, a=AA), color='pink', label='[B]')
+        # plt.plot(tdata, AA + AAB + BAAB, color='black', label='total')
+        # plt.legend()
+        # plt.show()
```
tBuLi: Should not be in the final version

tests/test_ode.py
```diff
@@ -97,6 +97,40 @@ def test_polgar(self):
...
+        ode_model = ODEModel(model_dict, initial={t: 0.0, a: a0, c: 0.0, d: 0.0})
+
+        # Generate some data
+        tdata = np.linspace(0, 3, 1000)
+        # Eval
+        AA, AAB, BAAB = ode_model(t=tdata, k=0.1, l=0.2, m=.3, p=0.3, a0=10)
+        fit = Fit(ode_model, t=tdata, a=AA, c=AAB, d=BAAB)
+        results = fit.execute()
+        print(results)
+        self.assertAlmostEqual(a0.value, 10)
```
tBuLi: I think the test should be a bit stronger. Perhaps give an initial guess of 11 instead of 10 to see if something really happened.

And add tests for `.params` and `.model_params` to see if they are what we expect them to be.

symfit/core/models.py
pckroon: It was needed because `a0` should not be passed to the python version of the component, to be integrated by scipy. Instead, `a0` should be passed to scipy in `initial_dependent`.
I'll add a comment

symfit/core/models.py
```diff
@@ -933,7 +933,7 @@ def __init__(self, model_dict, initial, *lsoda_args, **lsoda_kwargs):
         self.lsoda_args = lsoda_args
         self.lsoda_kwargs = lsoda_kwargs
 
-        sort_func = lambda symbol: str(symbol)
+        sort_func = str
```
tBuLi: good change, but I've sinse learned that it would be even more in keeping with sympy to use `lambda symbol: symbol.name`, because calling string is considered an anti-pattern whereas all symbols have a name. I've therefore started doing that in new code, so it might be worth updating this as well.

symfit/core/models.py
```diff
@@ -1095,12 +1108,17 @@ def eval_components(self, *args, **kwargs):
         # Properly ordered time axis containing t_initial
         t_total = np.concatenate((t_smaller[::-1][:-1], t_bigger))
 
+        # Call the numerical integrator. Note that we only pass the
+        # model_params, which will be used sympy_to_py to create something we
```
tBuLi: will be used *by*

symfit/core/models.py
```diff
@@ -1109,7 +1127,9 @@ def eval_components(self, *args, **kwargs):
             initial_dependent,
             t_smaller,
             args=tuple(
-                bound_arguments.arguments[param.name] for param in self.params),
+                bound_arguments.arguments[param.name]
+                for param in self.model_params
```
tBuLi: Personally I see this as an example of PEP8-ing gone mad. It's more readable if this was on one line, and the offense is not so big so I prefer it.

symfit/core/models.py
tBuLi: Seems less readable to me, is there a strong argument for preferring operator.attrgetter?

pckroon: It should be faster (I think). Readability is in part a matter of getting used to it. I personally prefer the attrgetter over dissecting a lambda.

tBuLi: I'd prefer the lambda, but it's not something I would stop the merging for so choose your favorite. I guess after this comment has been addressed we can merge this pr.

pckroon: My preference is the attrgetter, so squash and merge at your convenience.

JohnGoertz:
This is great! Is there a way to "scale" the initial values? For example, I have the following equation:

```python
logistic_eqn = {
    sf.D(s1,c): r*s1*(1-s1/K1)
}

logistic_model = sf.ODEModel(logistic_eqn, initial = {c:0.0,s1:s1_0})
```
This works fine if I give a good guess for s1_0, but I know that the local minima in the model are evenly distributed in the "log" space of the initial values. If I brute force the residuals, I find the global minima occurs at r = 0.8 and log10(s1_0) = -7.5, but there are other local minima at log10(s1_0) ~ -5.5, -6.5, -8.5, -9.5 with r = 0.6, 0.7, 0.9, and 1.0, respectively.

I worry that a naive minimizer will explore the initial values in linear space, thus inefficiently. I've tried giving a first guess for `s1_0` as -6.5 then specifying `initial = {...,s1:10**s1_0,...}`, but I get the error `can't convert expression to float`.

Any suggestions? I can provide more code if helpful, or post to StackOverflow if that's better.

pckroon:
Very interesting point!
Since it's an ODEModel we can't get the analytical Jacobian, and finding it with numerical differentiation in every step is too expensive. Some minimizers however (BFGS springs to mind) approximate it as they go along. Those *should* be able to pick up on the underlying structure. (Very few minimizers are truly naive).
For normal models you can work around this by introducing an intermediate variable, but I don't expect that will work here due to how this is implemented. Instead of passing just a Parameter as initial value, you kind of want to be able to pass an expression that evaluates to a single value.
Could you make an issue along those lines?

JohnGoertz:
Sure! See this issue: https://github.com/tBuLi/symfit/issues/284

PR #283, opened 2019-11-21
Update .travis.yml to add python 3.8

Jhsmit:


Jhsmit:
Python 3.8 requires the scipy < 1.2 requirement to be lifted. However, doing so breaks python 2.7 support. I'd say 3.8 > 2.7 and since python 2 support drops in about a month anyway its time to stop supporting py27.

pckroon:
Travis run 927.5 [1] looks as if the combination numpy-scipy is incompatible for Py2.7. This can probably be worked around by downgrading numpy to an appropriate version. I do think this is not worth the effort though.

[1] https://travis-ci.org/tBuLi/symfit/jobs/615081778?utm_medium=notification&utm_source=github_status

tBuLi:
I agree with dropping 2.7, it's time.

I do believe there was a good reason for freezing sympy <1.2. If memory serves me well it had something to do with code printing, but I can't find it back right now. But if the tests work with all the latest version then it should be fine?

PR #285, opened 2019-11-25
Verbose fit

### LINKED ISSUE #280: [Feature Request] Option to print progress updates
It would be great to have a 'verbose' option to report the progress of the fit, either through a progress bar a la [tqdm](https://tqdm.github.io/) or just by printing out the current iteration. I've been trying some very ODE parameter estimations, and it would be nice to have an estimate of how long is left.

pckroon:
Here's an initial mock-up of a `FitStatus` object that can be used to track convergence. I still need to add tests (and make sure the docs build).
@JohnGoertz would this suit your needs?

Closes #280


symfit/core/fit.py
```diff
@@ -568,19 +638,32 @@ def _init_constraints(self, constraints, model):
...
         return con_models
 
+    @keywordonly(callback=False)
     def execute(self, **minimize_options):
         """
         Execute the fit.
 
         :param minimize_options: keyword arguments to be passed to the specified
             minimizer.
+        :param callback: keyword only argument that will be propagated to the
```
tBuLi: This kwargs does not make sense to me. A callback has to be a callable. I think what you're going for is the keyword 'verbose'?

pckroon:
I also played around with caching the objective evaluations by having Fit.execute wrap self.objective.__call__ in a lru_cache, but that just plain doesn't work (discussion in #280). 
I'm completely happy to accept the performance hit one additional objective evaluation incurs per iteration, if someone explicitly asked for verbose feedback.

pckroon:
See also:
https://stackoverflow.com/questions/53423121/why-does-functools-lru-cache-not-cache-call-while-working-on-normal-methods

JohnGoertz:
That output looks great! Perfect information for slow fits. The "last" vs "avg" iter time will be particularly useful for diagnosing peculiarities. For example, I've recently been trying to fit large systems of ODEs that become "stiff" only for certain parameter values, resulting in massive slowdown. This would help catch that sort of thing.

symfit/core/fit.py
```diff
@@ -568,19 +636,47 @@ def _init_constraints(self, constraints, model):
...
+        elif callback:
+            minimize_options['callback'] = callback
         minimizer_ans = self.minimizer.execute(**minimize_options)
         minimizer_ans.covariance_matrix = self.covariance_matrix(
             dict(zip(self.model.params, minimizer_ans._popt))
         )
         # Overwrite the DummyModel with the current model
         minimizer_ans.model = self.model
         minimizer_ans.minimizer = self.minimizer
+        if minimize_options.get('callback', False):
```
tBuLi:
```python
if 'callback' in minimize_options:
```

symfit/core/fit.py
```diff
@@ -305,6 +306,73 @@ class HessApproximation(base, HessianObjectiveJacApprox):
...
+        :param params: 1D array-like with values for all non-fixed parameter
+            for this iteration. Used to call `objective`.
+        """
+        self.times.append(perf_counter())
+        free_params = dict(zip(map(str, self.objective.model.free_params), params))
+        obj_val = self.objective(**free_params)
+
+        # Add in fixed parameters
+        all_params = free_params.copy()
+        all_params.update({p: p.value for p in self.objective.model.params if p.fixed})
```
tBuLi: What is `all_params` used for?

symfit/core/fit.py
```diff
@@ -305,6 +306,73 @@ class HessApproximation(base, HessianObjectiveJacApprox):
...
+        """
+        Reports convergence statistics for this iteration, and stores that
+        information for later use.
+
+        :param params: 1D array-like with values for all non-fixed parameter
+            for this iteration. Used to call `objective`.
+        """
+        self.times.append(perf_counter())
+        free_params = dict(zip(map(str, self.objective.model.free_params), params))
+        obj_val = self.objective(**free_params)
```
tBuLi: I believe you can immediately do 
```python
obj_val = self.objective(params)
```

tBuLi:
I like this a lot too, just needs documentation and testing as far as I'm concerned. Other than that it's a good step in the right direction, and it also sets a good example for how people could define their own callback for their specific needs.

pckroon:
This should address your code issues, and adds tests. 
Where would you like the docs? I don't really see a "natural" place.

pckroon:
I'm getting unrelated test failures: HadamardProduct is not defined. Do you have any idea what could be causing them?

Issue #284, opened 2019-11-21
Scaling ODEModel initial value parameters


JohnGoertz:
Since 0.5.2 (and [this merge](https://github.com/tBuLi/symfit/pull/279#)) symfit can now estimate the initial values of ODEModels. However, there's no clear way to provide the scale of those initial values. For example, the following code fits a logistic curve to some data. Important information is the growth rate `r` and the initial signal `s1_0`. This works fine if I give a good guess for s1_0, but I know that the local minima in the model are evenly distributed in the "log" space of the initial values. If I brute force the residuals, I find the global minima occurs at r = 0.8 and log10(s1_0) = -7.5, but there are other local minima at log10(s1_0) ~ -5.5, -6.5, -8.5, -9.5 with r = 0.6, 0.7, 0.9, and 1.0, respectively.

Is there a way around this? Perhaps sympy could be used more explicitly, setting s1_0 as a relation to an intermediate value that lives in log-space?

```python
import numpy as np
import symfit as sf

signal1 = np.array([ 5.86039495e-03,  3.61153194e-03,  2.24330124e-03, -2.92966086e-03,
       -1.34148231e-03,  7.86443626e-04,  1.37743090e-03,  1.63747582e-05,
       -2.96159949e-03, -4.91545334e-03,  2.25744104e-03,  8.02806834e-03,
        6.86979013e-03, -3.99727245e-03, -5.43338189e-03, -4.08880449e-03,
        2.11456048e-03,  1.97120626e-02,  5.00087356e-02,  1.14839342e-01,
        2.26256041e-01,  3.75486015e-01,  5.36502495e-01,  6.94936368e-01,
        8.08767694e-01,  8.73566838e-01,  9.04374323e-01,  9.18962105e-01,
        9.33002509e-01,  9.26539485e-01,  9.22790406e-01,  9.27219537e-01,
        9.30125574e-01,  9.37782682e-01,  9.31860668e-01,  9.25248598e-01,
        9.27760548e-01,  9.21753351e-01,  9.28642494e-01,  9.33063441e-01,
        9.26756334e-01,  9.24954549e-01,  9.24742853e-01,  9.21482340e-01,
        9.26562524e-01,  9.23340814e-01,  9.32431892e-01,  9.27025424e-01,
        9.18785776e-01,  9.17351300e-01,  9.22062558e-01,  9.27517326e-01,
        9.25602973e-01,  9.27516517e-01,  9.26961663e-01,  9.31088965e-01,
[...]

PR #287, opened 2019-12-04
Make constraints as Model instead of the same as their parent model

### LINKED ISSUE #282: ODEModel with constraints gives error
Adding constraints to the fitting of an ODEModel raises an error, `__init__() missing 1 required positional argument: 'initial'`

Taking the simple coupled ODE model from the docs and adding an equality constraint to the variables (one that will necessarily be true anyways) gives the following output:

```python
tdata = np.array([10, 26, 44, 70, 120])
adata = 10e-4 * np.array([44, 34, 27, 20, 14])
a, b, t = sf.variables('a, b, t')
k = sf.Parameter('k', 0.1)
a0 = 54 * 10e-4

model_dict = {
    sf.D(a, t): - k * a**2,
    sf.D(b, t): k * a**2,
}

ode_model = sf.ODEModel(model_dict, initial={t: 0.0, a: a0, b: 0.0})

constraints = [sf.Eq(a+b,0.055)]

fit = sf.Fit(ode_model, t=tdata, a=adata, b=None, constraints=constraints)
fit_result = fit.execute()
print(fit_result)
```

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-133-bcdec22720dd> in <module>
     14 constraints = [sf.Eq(a+b,0.055)]
     15 
---> 16 fit = sf.Fit(ode_model, t=tdata, a=adata, b=None, constraints=constraints)
     17 fit_result = fit.execute()
     18 print(fit_result)

C:\ProgramData\Anaconda3\envs\sf_env\lib\site-packages\symfit\core\support.py in wrapped_func(*args, **kwargs)
    421                     else:
    422                         bound_args.arguments[param.name] = param.default
--> 423             return func(*bound_args.args, **bound_args.kwargs)
    424         return wrapped_func
    425 

C:\ProgramData\Anaconda3\envs\sf_env\lib\site-packages\symfit\core\fit.py in __init__(self, model, *ordered_data, **named_data)
    374 
    375         self.constraints = self._init_constraints(constraints=constraints,
--> 376                                                   model=self.model)
    377 
    378         # Bind as much as possible the provided arguments.

C:\ProgramData\Anaconda3\envs\sf_env\lib\site-packages\symfit\core\fit.py in _init_constraints(self, constraints, model)
    565                 else:
    566                     con_models.append(
--> 567                         model.__class__.as_constraint(constraint, model)
    568                     )
    569         return con_models

C:\ProgramData\Anaconda3\envs\sf_env\lib\site-packages\symfit\core\models.py in as_constraint(cls, constraint, model, constraint_type, **init_kwargs)
    157         instance = cls.with_dependencies(constraint,
    158                                          dependency_model=model,
--> 159                                          **init_kwargs)
    160 
    161         # Check if the constraint_type is allowed, and flip the sign if needed

C:\ProgramData\Anaconda3\envs\sf_env\lib\site-packages\symfit\core\models.py in with_dependencies(cls, model_expr, dependency_model, **init_kwargs)
    207         :return: A stand-alone :class:`~symfit.core.models.BaseModel` subclass.
    208         """
--> 209         model = cls(model_expr, **init_kwargs)  # Initiate model instance.
    210         if any(var in dependency_model for var in model.independent_vars):
    211             # This model depends on the output of the dependency_model,

TypeError: __init__() missing 1 required positional argument: 'initial'
```

pckroon:
ODEModels make for poor constraints, since they require an `initial` argument, which is not passed by `BaseModel.as_constraint`. This does mean that if you want to use an ODEModel as constraint, you need to initialize your constraint model yourself, and pass it.
On top of that, this PR adds initial ODE parameters to the `connectivity_mapping`.

Fixes #282


tests/test_ode.py
```diff
@@ -129,6 +129,48 @@ def test_initial_parameters(self):
...
+        ode_model = ODEModel(model_dict, initial={t: 0.0, a: a0, c: c0, d: 0.0})
+
+        # Generate some data
+        tdata = np.linspace(0, 3, 1000)
+        # Eval
+        AA, AAB, BAAB = ode_model(t=tdata, k=0.1, l=0.2, m=.3, p=0.3, a0=10, c0=0)
+
+        constraints = [Ge(a0, c0)]
+
+        fit = Fit(ode_model, t=tdata, a=AA, c=AAB, d=BAAB)
```
tBuLi: The constraints are not passed to `Fit`. It is probably better to use an equatilty or make the constraint the other way around, since initially the constraint is already true so we don't know if it was enforced properly.

tBuLi:
Nice one. But I think the original reason for using the same class is that when adding constraints to a `CallableNumericalModel`, the constraint also needs to be of that same type, in case the constraint reuses some of the components of the original model. This functionality is now broken.

E.g.
```python
model = CallableNumericalModel({y: lambda x: a * x**2}, connectivity_mapping={y: {x, a}})
constraint = CallableNumericalModel.as_constraint({z: lambda y: np.sum(y)}, connectivity_mapping={z: {y}}, model=model, constraint_type=Eq)
```

I'm actually surprised the tests didn't fail for this, I guess that means I never added proper testing for the above functionality. Perhaps whoever of us has time first can check if this indeed works, because this is very nice functionality to have.

So I think I'm more in favor of adding a simple check for ODEModels and only do your new fix there but keep the standard behavior otherwise.

tBuLi:
I also think that ultimately the solution would be to rework `ODEModel` into a subclass of `CallableNumericalModel` so we can also more easily mix component of ODE and non-ODE type like in the recent issue #284.

pckroon:
> I also think that ultimately the solution would be to rework ODEModel into a subclass of CallableNumericalModel so we can also more easily mix component of ODE and non-ODE type like in the recent issue #284, but might be bigger than the current issue.
I forgot about this TODO. I'll fix both in one go.

tBuLi:
Btw, a potential unified syntax could be as follows:
```python
ODEModel({
    D(y, t): - k * y,
    y: 1,
    t: 0
})
```
This would move everything to the model_dict and makes the ambition of having expressions as initial conditions much easier. It would also streamline internal code regarding ODEModels, though at the expense of some readability for the user I think, since it is not immediately obvious that these are to be read as initial conditions. What do you think?

tBuLi:
It could also be a good idea an improved version of the current ODEModel but call it RawODEModel instead, and then make a new ODEModel which allows for mixing of ODE and normal expressions, but which then seperates the ODE components into a RawODEModel and then adds this as a numerical component to a CallableNumericalModel. This would essentially automate the process here: https://symfit.readthedocs.io/en/stable/examples/ex_CallableNumericalModel_ode.html

PR #296, opened 2020-02-06
Fixes issue 295

Jhsmit:
Checks if any bounds are none before comparing min/min

For some reason there are some commits listed which are already on symfit master

### DIFF
--- a/symfit/core/support.py
+++ b/symfit/core/support.py
@@ -230,7 +230,9 @@ def parameters(names, **kwargs):
 
     if 'min' in sequences and 'max' in sequences:
         for min, max in zip(sequences['min'], sequences['max']):
-            if min > max:
+            if min is None or max is None:
+                pass
+            elif min > max:
                 raise ValueError('The value of `min` should be less than or'
                                  ' equal to the value of `max`.')
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,6 @@ jobs:
     - python: "3.6"
     - python: "3.7"
     - python: "3.8"
-    - python: "2.7"
 
     - stage: docs
       python: "3.7"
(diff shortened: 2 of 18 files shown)


symfit/core/support.py
```diff
@@ -230,7 +230,9 @@ def parameters(names, **kwargs):
 
     if 'min' in sequences and 'max' in sequences:
         for min, max in zip(sequences['min'], sequences['max']):
-            if min > max:
+            if min is None or max is None:
+                pass
+            elif min > max:
                 raise ValueError('The value of `min` should be less than or'
                                  ' equal to the value of `max`.')
```
pckroon: Could you condense these conditions? `min is not None and max is not None and min > max`?
Alternatively, completely remove the check here, and make min/max a property of Parameter, and do the checking in the setter. That should clean up some logic around the place.

Jhsmit:
Edited on GitHub which also added a newline at the end of the file

pckroon:
Ok, but then you also need to add the Parameter.min and Parameter.max properties... Currently we'd just lose a useful check instead of moving it.

Jhsmit:
The current check I've removed was redundant because its also in the ``__init__`` of ``Parameter``, where also the possibility of ``None`` is checked in one line.

In any case, the getter/setter would still add functionality because it prevents users from setting faulty limits after initialization of the parameters.

pckroon:
`parameters` does not call `Parameter.__init__`. Instead it does `for param, value in zip(params, values): (param, key, value)` (https://github.com/tBuLi/symfit/pull/296/files#diff-f73c0cdf32475d7606d6d487cae94340R245).

Of course, once there's setter methods the check in `Parameter.__init__` will become redundant.

Jhsmit:
Ah I see, ``Parameter.__init__`` does get called but not with the user supplied min/max values which indeed are set later.
I'll put in the setter methods instead.

tBuLi:
Yeah I agree that putting this check in the setter is probably a nicer solution. Another solution could be to initiate each object properly and not set it after the fact.

Issue #290, opened 2020-01-06
Allow declaration of Dummy arguments instead of just Symbols


JohnGoertz:
I'm not sure if this would break everything, but it would be very useful to be able to construct parameters/variables that inherit sympy's `Dummy` class instead of the `Symbol` class. The reason being that if multiple variables are created with the same string input, they are *the same object* and changing one changes the other.

```python
import symfit as sf
symbols = 'x,y'
a = sf.parameters(symbols)
b = sf.parameters(symbols)
print(a is b)
print(a[1] is b[1])
print(b[1].value)
a[1].value = 2
print(b[1].value)
```
```
False
True
1.0
2
```
This is even true if these variables are containerized into class instances.

```python
class MyContainer:
    def __init__(self,symbols):
        self.params = sf.parameters(symbols)
        
a = MyContainer(symbols)
b = MyContainer(symbols)
print(a is b)
print(a.params[1] is b.params[1])
print(b.params[1].value)
a.params[1].value = 2
print(b.params[1].value)
```
```
False
True
1.0
2
```

Sympy solves this through the Dummy variable class, so if there was an alternative constructor for symfit's `Argument` class to extend Dummy, that (should) fix the issue.

```python
import sympy
a = sympy.symbols(symbols)
b = sympy.symbols(symbols)
print(a is b)
print(a[1] is b[1])
a = sympy.symbols(symbols, cls=sympy.Dummy)
b = sympy.symbols(symbols, cls=sympy.Dummy)
print(a is b)
print(a[1] is b[1])
```
```
False
True
False
False
```

Now, I'm not sure if this is more properly a sympy bug or if this is actually desirable behavior, or if it would be a nightmare to implement, but... it would be nice.

pckroon:
I think that it's desired behaviour, as far as sympy is concerned. The good news is that we subclass Symbol fairly transparently: could you try `b = symfit.Parameter(name='b', cls=sympy.Dummy)` and see where it breaks?

Jhsmit:
This was also discussed a while ago in #124

Jhsmit:
@pckroon this breaks directly on ``TypeError: __new__() got multiple values for argument 'cls'``

Its possible to change the parameter creation from 
```python
    params = symbols(names, cls=Parameter, seq=True, **kwargs)
```
```python
    cls = kwargs.pop('cls', Parameter)
    params = symbols(names, cls=cls, seq=True, **kwargs)
```

[...]

PR #298, opened 2020-02-07
Added metaclass to create dummy Paramter/Variables

Jhsmit:
This is a bit of a hack, but it allows users to force creation of dummy Parameter / Variables instead of symbol ones (see #290).

Usage:
```python
import builtins
builtins.symfit_dummy = True

from symfit import Parameter

a = Parameter('x', value=2)
b = Parameter('x', value=3)

print(a.value, b.value)


>>> (2, 3)
```

Where in normal ``symfit`` behaviour the result would be ``(3, 3)``.

Why would you want to use this? Well, if your application of ``symfit`` is one script, one model kind of fitting then there is probably very little use. But if you make a bigger application where many models are created and active at the same time (thread safe?) then dummy behaviour is desired.

Also, lets say you happen to be making your parmeters like this:

```python
A = Parameter()
sig = Parameter('par_0')
```
Without dummies, these two parameters are the same, and fitting fails without an error. With dummies, it still doesnt work, but at least you get an error :)

Yes, I agree the implementation is a bit sketchy and could be considered dark magic. Suggestions are welcome. Maybe some split of the inheritance into ``Parameter`` and ``DummyParameter``, but because of all the implementations of ``__new__`` it seems that ``Dummy`` somehow needs to be introduced just after ``Argument`` in the mro. 
The hack via ``builtins`` is needed because the information on wheter to dummy or not needs to be passed before any symfit imports as that is when the classes are created.
``builtins`` is py3 only (did we drop py2 already?)

### DIFF
--- a/symfit/core/argument.py
+++ b/symfit/core/argument.py
@@ -2,10 +2,24 @@ from collections import defaultdict
 import numbers
 import warnings
 
-from sympy.core.symbol import Symbol
+from sympy.core.symbol import Symbol, Dummy
+from sympy.core.assumptions import ManagedProperties
+import builtins
 
 
-class Argument(Symbol):
+class DummyMeta(ManagedProperties):
+
+    def __new__(cls, name, bases, namespace, **kwargs):
+        try:
+            if name == 'Argument' and builtins.symfit_dummy:
+                bases = (Dummy, )
+        except AttributeError:
+            pass
+
+        cls = super().__new__(cls, name, bases, namespace)
+        return cls
+
+class Argument(Symbol, metaclass=DummyMeta):
     """
     Base class for :mod:`symfit` symbols. This helps make :mod:`symfit` symbols
     distinguishable from :mod:`sympy` symbols.


tBuLi:
I like the idea for creating this, but I would rather choose a solution that stays closer to sympy to prevent problems in the future. I think the following object structure should work:

BaseParameter - defines slots min, max, fixed, value. This is a mixin class and doesn't inherit from anything.
Parameter(BaseParameter, Argument)
DummyParameter(BaseParameter, sympy.Dummy)

I think this should work, what about you?

Jhsmit:
Yes I agree this PR should not be merged under any circumstances. But for it a was a fun trip into the madness that is symfit, sympy and metaclasses. 

The structure you proposed is indeed the best solution in my opinion. I dont remember the details but I had trouble envisioning it because of I ran into some MRO issues. But as a mixin it might work. 

We'd also need to do the same thing for Variables then but that should be much easier.

tBuLi:
Yeah I had those same problems when trying to include indexed variables and parameters and back then I eventually found that a mixin was the only way of getting it to work if you didn't want to go down a very dark path.

I also don't think the solution above should be very hard to implement, and I would prefer having DummyParameters and DummyVariables as seperate classes so we don't create unexpected problems. And we can add a `dummy` keyword to the `parameters` and `variables` support functions for convenience.

Jhsmit:
Perhaps Parameters, Variables should always be Dummy's in a future (major) version of symfit?
I don't see a clear use case for the non-dummy behavior. 

Perhaps you could consider someone wanting to do this:

```python
model = {
Variable('y1'): Parameter('a')*Variable('x') + Parameter('b'),
Variable('y2'): Parameter('a')*Variable('x')**2
}
```

Which with dummy variables you would have to define by:

```python
y1_var = Variable('y1')
y2_var = Variable('y2')
a_par = Parameter('a')
b_par = Parameter('b')

model = { ...
```
Where I do agree that the former has a nicer syntax and the latter pollutes the local scope but this example ignores the fact that on parameters you have to define initial guesses so usually you cannot write your problem as option 1 anyway.

PR #300, opened 2020-03-18
Added HadamardPower printing

tBuLi:
Fix the error due to HadamardPower not being printed yet for new sympy versions.

### DIFF
--- a/symfit/core/printing.py
+++ b/symfit/core/printing.py
@@ -6,7 +6,7 @@ slightly different behavior from the standard one.
 
 Users using both ``symfit`` and ``sympy`` should be aware of this.
 """
-import sympy
+import pkg_resources
 from sympy import HadamardProduct, MatPow, Idx, Inverse
 from sympy.printing.codeprinter import CodePrinter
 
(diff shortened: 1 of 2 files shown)


symfit/core/printing.py
```diff
@@ -41,6 +41,14 @@ def _print_HadamardProduct(self, printer):
                       printer.doprint(self.args[1]))
 HadamardProduct._numpycode = _print_HadamardProduct
 
+if tuple(int(i) for i in sympy.__version__.split('.')) > (1,4):
```
pckroon: Isn't there some more canonical way of doing this in e.g. setuptools, distutils, or pks_resources?

tBuLi: Interesting point, but reading about it now I'm not so sure everyone has pks_resources. It seems the `packaging` package is good for doing the comparisons, but that would mean an extra dependency.
https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python/21065570

Maybe add packaging as a dependency because it is light-weight and use sympy.__version__ to get the version number?

pckroon: `pkg_resources` comes with setuptools, so effectively everyone has that. `pkg_resources.parse_version()` returns something that's comparable. https://setuptools.readthedocs.io/en/latest/pkg_resources.html#parsing-utilities

It's up to you whether it's worth an additional dependency.

tBuLi: Yeah I used pkg_resources as well, I do agree that it is nicer to use a canonical solution.

PR #316, opened 2020-09-10
Update requirements.txt

Jhsmit:
Addded setuptools (pkg_resources) requirement and removed py2 requirements

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,5 @@ numpy >= 1.12
 scipy >= 1.0
 sympy >= 1.2
 toposort
-funcsigs; python_version < '3.0'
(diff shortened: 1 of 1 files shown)


pckroon:
Setuptools and pbr are already listed as requirements in https://github.com/tBuLi/symfit/blob/master/setup.py#L4. Why is this needed?

Jhsmit:
This is because setuptools is not only needed for setup, but also runtime. 
Its used in ``__init__`` to set ``__version__`` as well as in printing: https://github.com/tBuLi/symfit/blob/7399494caac1b4e48611ca8c71a088c34544f737/symfit/core/printing.py#L9

pckroon (approved):

PR #319, opened 2020-10-21
Pickle ModelOutput

pckroon:
Turns out ModelOutput objects were not unpicklable. This PR fixes that.

Without the `__getnewargs_ex__` method a missing positional argument error is raised on `__new__`, without the get/setstate methods you get an infinite recursion loop.


symfit/core/models.py
```diff
@@ -60,11 +60,20 @@ def __init__(self, variables, output):
         :param output: The output of a call which should be mapped to
             ``variables``.
         """
-        self.variables = variables
+        self.variables = list(variables)
         self.output = output
         self.output_dict = OrderedDict(zip(variables, output))
         self.variable_names = {var.name: var for var in variables}
 
+    def __getnewargs_ex__(self):
```
tBuLi, line 68: Is it possible to define __getnewargs__ instead and return args instead of kwargs? That should fix the failing tests for py3.5?

pckroon:
Done, but note that py3.5 is EOL.

tBuLi (approved):
Looks good to me, go ahead and merge if you are also satisfied.

PR #321, opened 2020-10-30
Change license to MIT

### LINKED ISSUE #309: License clarifications
Hi again,
so symfit got rejected from Debian due to some license unclarities.
So the first thing is that
https://github.com/tBuLi/symfit/blob/ed20d296a818f6e8ce89efae4e9f3dce635ff4db/setup.cfg#L8
states an MIT license which seems like a copy-and-paste mistake.

The second thing is GPL2 clarification. The license file is GPLv2, but I would like to know if the `or-later` stuff applies.

My recommendation is to just put `GPL-2.0-only` or `GPL-2.0-or-later` (which one doesn't really matter to me) in `setup.cfg`. Those are SPDX identifiers (https://spdx.org/licenses/), the difference becomes clear when you compare the Standard License Headers.

pckroon:
This PR changes the license to MIT, as discussed in #309, and adds SPDX license headers to all py files (excluding the examples). 
I didn't add the license header to `core/leastsqbound.py`, I seem to remember we borrowed that code from elsewhere. @tBuLi do you remember?

Closes #309


tBuLi:
Correct, leastsqbounds is a direct copy paste of a repo with the same name. (There should also be an open issue to remove this since this functionality has since been included in scipy, but that’s another PR)

tBuLi:
Might be a noob question, but what are these SPDX headers for?

pckroon:
> Correct, leastsqbounds is a direct copy paste of a repo with the same name. (There should also be an open issue to remove this since this functionality has since been included in scipy, but that’s another PR)

I'll remove that file in this PR then, since it does have implications for the license.

> Might be a noob question, but what are these SPDX headers for?

Every file needs a header stating under what license it can be distributed. As I understand the SPDX format is just a short, standardized way of doing that. @stephanlachnit?

stephanlachnit:
> Might be a noob question, but what are these SPDX headers for?

> Every file needs a header stating under what license it can be distributed. As I understand the SPDX format is just a short, standardized way of doing that. @stephanlachnit?

Yes exactly. It's a standard from the Linux Foundation (slightly changed from the Free Software Foundation Europe to standardize the copyright owner) to easily parse files for their copyright.

With this standard it's easy to import files from other repos and still respect their copyright information, also you can quickly create copyright reports, which might be usefull for example for distributors like Debian.

stephanlachnit:
@tBuLi is there any chance we can get this merged with a new release before February? Else it won't make it for the Debain 11 freeze starting on February 12th.

pckroon:
@stephanlachnit Sorry for the long wait and the missed deadline! Maybe we can try again with Debian 12?

stephanlachnit:
Jup, no problem. I will package it anyway a soon a there is a ne release.

pckroon:
Awesome thanks! Will you get notified automatically once there's a new release, or should I @ you somewhere once we mint a new release?

stephanlachnit:
I'm subscribed to the GitHub releases, so as long as they will still be done I will get notified.

stephanlachnit:
Are there any plans to release a new version? Debian 11 has released now and new software can be added, and I would love to add symfit.

PR #328, opened 2021-03-29
Fixed interactive fit tests, added ODE test

Jhsmit:
Changed check of title from `x.name` to `x` (as it is formatted in the code)
Added test from ODE2D example


pckroon (approved):
Many thanks, look good to me!

For future PRs, could you split cases like this one into multiple commits? I.e. one that fixes the latexification of x, and one that adds the new testcase?

Jhsmit:
Yeah, good point, I wanted to stage them in chunks but they were overlapping.
My git etiquette is terrible :P

pckroon:
Live and learn :)

PR #329, opened 2021-03-29
CI with GitHub Actions

Jhsmit:
Moved CI to GitHub actions.

Currently functional:
- [ ] pytest with coverage
- [ ] Upload of coverage to Coveralls
- [ ] Deploy on PyPi
- [x] Build docs

First 3 should work but cannot be tested from this PR

### DIFF
--- a/.github/workflows/deploy_pypi.yml
+++ b/.github/workflows/deploy_pypi.yml
@@ -1,4 +1,4 @@
-name: PyPi distribute main release
+name: PyPi deploy release
 on: [release]
 
 jobs:
--- a/.github/workflows/deploy_pypi.yml
+++ b/.github/workflows/deploy_pypi.yml
@@ -1,0 +1,26 @@
+name: PyPi distribute main release
+on: [release]
+
+jobs:
+  build-n-publish:
+    name: Build and publish Python 🐍 distributions 📦 to PyPI
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Fetch all history for all tags and branches
+      run: git fetch --prune --unshallow
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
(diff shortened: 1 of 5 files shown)


Jhsmit:
@tBuLi @pckroon For the PyPi deploy to work we would need a PyPI token to be added to GitHub's secrets:
https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#saving-credentials-on-github

See also:
https://docs.github.com/en/actions/reference/encrypted-secrets

Jhsmit:
The current Travis CI procedure also builds the docs. Why is this needed? Arent they built by Readthedocs? Or is it to test if building the docs actually works?

pckroon:
> The current Travis CI procedure also builds the docs. Why is this needed? Arent they built by Readthedocs? Or is it to test if building the docs actually works?

Indeed. The CI should try to build the docs in nitpick mode to make sure it works, and RTD builds them without the nitpick flag so it's more likely to just work.

Jhsmit:
Building docs was added and works.
Upload to PyPi done but currently fails due to missing tokens
Pytests run an pass for ubuntu/mac, fails on windows, but is fixed by #331

tBuLi:
First of all, excellent PR!

Alright, there are now PYPI_API_TOKEN, TEST_PYPI_API_TOKEN , and COVERALLS_REPO_TOKEN, so at least the pypi ones should now work. 

Why does the coveralls code use GITHUB_TOKEN? If I understood correctly, shouldn't that be COVERALLS_REPO_TOKEN?

Jhsmit:
The coveralls GitHub action I've taken from the coveralls-pyton docs: https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html#github-actions-support

Where it says that the COVERALLS_REPO_TOKEN is not needed.

I think this is ready to merge now, although I think #331 should be merged first such that when this merges the tests should pass and we can see if upload to coveralls works

PR #336, opened 2021-09-21
Added no coveralls pytest action

Jhsmit:
Temporary workaround coveralls downtime


tBuLi:
Note: this got advanced from a temporary fix to a full split to prevent this issue in the future. If coveralls is down that shouldn't prevent us from testing our code.

Issue #339, opened 2021-10-11
ValueError for fit of ODEModel with additional data for t0


MrAnimaniac:
I might have found a bug likely associate with `simfit.ODEModel` combined with fitting. More precisely, I assume it might be connected to `ODEModel`'s the `initial` -dict. The issue can be reproduced using the first example provided in the simfit docs, “4.5 ODE Fitting”. 
```
from symfit import Parameter, variables, Fit, D, ODEModel
import matplotlib.pyplot as plt
import numpy as np


tdata = np.array([0, 0, 10, 26, 44, 70, 120])
adata = 10e-4 * np.array([54, 54, 44, 34, 27, 20, 14])


a, b, t = variables('a, b, t')
k = Parameter('k', 0.1)

model_dict = {
    D(a, t): - k * a**2,
    D(b, t): k * a**2,
}

ode_model = ODEModel(model_dict, initial={t: tdata[0], a: adata[0], b: 0.0})

fit = Fit(ode_model, t=tdata, a=adata, b=None)
fit_result = fit.execute()

# Generate some data
tvec = np.linspace(0, 500, 1000)

A, B = ode_model(t=tvec, **fit_result.params)
plt.plot(tvec, A, label='[A]')
plt.plot(tvec, B, label='[B]')
plt.scatter(tdata, adata)
plt.legend()
plt.show()
```

will raise following error: 
`ValueError: operands could not be broadcast together with remapped shapes [original->remapped]: (8,)  and requested shape (7,)`

However, it will behave as expected if you change ` tdata = np.array([0, 10, 10, 26, 44, 70, 120])`. In this example, `adata[0]` will be repeatedly handled over to the algorithm. This issue, however, rises independently. 

Furthermore, the difference between original (m) and requested shape (n) mentioned in the error message is proportional to the frequency of repeated t0-values (f): _m - n = f-1_.

PR #343, opened 2021-10-28
Switch out odeint for the more modern solve_ivp

pckroon:
This PR switches the ODEModel from the old `odeint` to the newer `solve_ivp`. And there was much rejoicing. Also does some minor polishing left and right.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,10 +1,10 @@
 [metadata]
 name = symfit
 summary = Symbolic Fitting; fitting as it should be.
-description-file = README.rst
+description_file = README.rst
 author = Martin Roelfs
 author_email = martin.roelfs@kuleuven.be
-home-page = https://github.com/tBuLi/symfit
+home_page = https://github.com/tBuLi/symfit
 license = MIT
 classifier =
     Development Status :: 4 - Beta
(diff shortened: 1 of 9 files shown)


tBuLi:
Almost all failing tests have been resolved by setting the `atol/rtol` LSODA settings back to the original settings of `odeint`. (I've also rebased on he current master branch) Furthermore I made some performance improvements by not repeating the generation of the system of equations every iteration, but instead turned that into a `cached_property`.

The range also does not seem to need `+- 1e-8` or something similar, works just fine for all our tests without it (and according to the documentation) so not sure why this was needed?

There are only two tests left that fail: pickling and parameters as initial guesses. Pickling seems to be simple to fix: this is due to the fact that we no longer support positional arguments for the  ODE solver, which seems like a good thing to me so we can just change the test. The latter test is a bit more complicated: LSODA throws an error which seems to be due to loss of precission. But this is probably also solvable by identifying the changed standard settings as compared to `odeint` which lead to this problem.

pckroon:
> Almost all failing tests have been resolved by setting the atol/rtol LSODA settings back to the original settings of odeint. (I've also rebased on he current master branch) Furthermore I made some performance improvements by not repeating the generation of the system of equations every iteration, but instead turned that into a cached_property.

Good finds :)

> The range also does not seem to need +- 1e-8 or something similar, works just fine for all our tests without it (and according to the documentation) so not sure why this was needed?

This used to break when I tried it (I was playing with #339 at the time). Maybe scipy got updated in the meantime?

Issue #344, opened 2021-11-04
Fix deprecated syntax in setup.cfg


ford--prefect:
When building, I receive the following warnings:

````
/usr/lib/python3.9/site-packages/setuptools/dist.py:697: UserWarning: Usage of dash-separated 'description-file' will not be supported in future versions. Please use the underscore name 'description_file' instead
/usr/lib/python3.9/site-packages/setuptools/dist.py:697: UserWarning: Usage of dash-separated 'home-page' will not be supported in future versions. Please use the underscore name 'home_page' instead
````

Sounds simple enough to quickly fix.
Thanks!

ford--prefect:
Close, fixed with #345.

PR #345, opened 2021-11-04
fix deprecated syntax in setup.cfg (#344)

ford--prefect:
Fix for issue #344


pckroon (approved):

pckroon:
Looks good to me! Thanks for the help! :)

pckroon:
I don't see how your change can affect the test outcomes, so I'll assume those are fixed in #342

PR #354, opened 2022-10-25
Update build matrix to py 3.7-3.11. Py3.6 is EOL

pckroon:


tBuLi:
Thanks for this PR, I like it! Quick question, why the switch to `pytest-cov`? Just curious about the reason, but other than that, please merge it if you consider this ready.

pckroon:
Besides "because it works", it also does some intelligent things regarding e.g. multiprocessing and coverage. In general it's just a good idea.

PR #365, opened 2023-01-24
Py3 final migration

tBuLi:
There were a lot of fossils in the code from the time we still had to support both py27 and py3x. Those days are long gone, so it's time for a major clean-up. The key points of this PR are:

- Keyword-only arguments can just be used in the way Python intended. There is no longer a need for the `keywordonly` decorator.
- Remove all explicit `sys.version_info >= (3, 0)` checks, since the answer is always `True`.
- Use the built-in partial and delete our own implementation which mimicked the py3 partial function in py27.

Up for debate:
- Python has a built-in `cached_property` since python 3.8. However, if I swap our implementation for this, pickling breaks in a way that is not that is not obvious to fix. So not sure yet if this is worth the effort. The problem is that now we simply check if the class dict contains entries starting with `_cached` because those are, well, cached. But this no longer works with the built-in, so we would have to get smart, which is difficult.


pckroon (approved):
Beautiful. I love this type of PR.
About `cached_property`, could you open an issue? Definitely worth looking into, but not urgent.

Did you figure out what's wrong with the mac tests? In general I think we can drastically trim our test matrix. For example, only run py 3.7-3.11 on linux, and just 3.11 on mac.

tBuLi:
Yeah I agree, we can leave `cached_property` as an issue.

Weirdly enough I am unable to reproduce the failing test for `test_global_opt.py::TestGlobalOptGaussian::test_chained_min` locally, even though I recreated the same environment on my own mac. Weirdly enough however, the error is very consistent in all macos tests in our build matrix it seems. Not sure what to do from here.

I agree on thinning out the build matrix btw, it seems a bit excessive.

PR #366, opened 2023-03-07
test_chained_min: relax test constraints

stephanlachnit:
This test fails on ppc64el otherwise, see e.g. https://ci.debian.net/data/autopkgtest/testing/ppc64el/s/symfit/32008561/log.gz (search for `test_chained_min`).

### DIFF
--- a/tests/test_global_opt.py
+++ b/tests/test_global_opt.py
@@ -82,8 +82,8 @@ class TestGlobalOptGaussian:
         fit_result = fit.execute(
             DifferentialEvolution={'seed': 0, 'tol': 1e-4, 'maxiter': 10}
         )
-        assert fit_result.value(self.x0_1) == pytest.approx(0.4, 1e-4)
-        assert fit_result.value(self.y0_1) == pytest.approx(0.4, 1e-4)
+        assert fit_result.value(self.x0_1) == pytest.approx(0.4, 1e-2)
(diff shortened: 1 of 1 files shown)


pckroon:
Thanks. I have no issues with this at all. @tBuLi any concerns?

I would really appreciate your input on how to deal with these type numerical testing issues though, at a more general level. Does it make sense to just relax them until it works, or does that simply invalidate the tests?

pckroon (approved):

pckroon:
> I guess it boils down to: do you except the tests to show that it works in general?

Great question :)
We (I at least) do *not* want to test the scipy optimizers. I do want to test the symfit side of code. I guess the summary is that our current tests are rather integrative in nature, which is... not ideal. 
My vote is that this change is OK; and we need to revisit our tests (but who has the time...)

@tBuLi Please merge and release when able :)

tBuLi:
I agree that we should only be testing our symfit code, not scipy. Unfortunately this is a bit hard to untangle, and there are a lot of tests that are indeed kind of written as performance tests even though their goal is just to make sure our code is executed correctly.

So I have no problem with this PR, improving the tests one step at a time ;). I don't think we need a new release just for this though, or do we?

stephanlachnit:
> So I have no problem with this PR, improving the tests one step at a time ;). I don't think we need a new release just for this though, or do we?

Nah that's fine, thx!

PR #370, opened 2023-09-18
First version of `symmip`, which adds tentative MIP support

tBuLi:
First version of `symmip`, a symbolic wrapper around MIP solvers. Currently only support the Gurobi API, but the code is setup in such a way that we can easily add other solvers.

Added as a seperate module, because merging it into the existing symfit infrastructure was less than obvious. In particular, the existing Models, objectives and minimizers are only very expensive boilerplate with no added value for MIPs, so for now I cut ties with this formalism. We can always still plug this MIP solver back into the Fit API if we have a good MIP API to begin with.


lesshaste:
Highs would be wonderful to add as it is open source.

examples/mip/bilinear.py
```diff
@@ -0,0 +1,28 @@
...
+objective = 1.0 * x
+constraints = [
+    x + y + z <= 10,
+    x*y <= 2,
+    Eq(x*z + y*z, 1),
+]
+
+mip = MIP(- objective, constraints=constraints)
+# For full control, we still need to access the model directly.
+mip.backend.model.Params.NonConvex = 2
```
pckroon: Oh I *don't* like this at all...

examples/mip/bilinear.py
```diff
@@ -0,0 +1,28 @@
...
+mip_result = mip.execute()
+
+print(mip_result)
```
pckroon, line 26: Missing empty line

examples/mip/bilinear.py
```diff
@@ -0,0 +1,28 @@
+# Inspired by https://www.gurobi.com/documentation/9.5/examples/bilinear_py.html#subsubsection:bilinear.py
+#
+# This example formulates and solves the following simple bilinear model:
+#  maximize    x
+#  subject to  x + y + z <= 10
+#              x * y <= 2         (bilinear inequality)
+#              x * z + y * z = 1  (bilinear equality)
+#              x, y, z non-negative (x integral in second version)
```
pckroon, line 8: For my understanding, this is not a MIP, but just an almost-linear problem? In other words, there are no integer constraints. In other other words, you could solve this with the existing minimizers

pckroon: Second note, I don't see the non-negative boundaries in the code below

examples/mip/mip1.py
```diff
@@ -0,0 +1,32 @@
...
+
+x, y, z = parameters('x, y, z', binary=True)
+
+objective = x + y + 2 * z
+constraints = [
+    x + 2 * y + 3 * z <= 4,
+    x + y >= 1
+]
+
+fit = MIP(- objective, constraints=constraints, backend=GurobiBackend)
```
pckroon: Why explicit backend selection here?

examples/mip/mip1.py
```diff
@@ -0,0 +1,32 @@
...
+fit_result = fit.execute()
+
+print(f"Optimal objective value: {fit_result.objective_value}")
+print(
+    f"Solution values: "
+    f"x={fit_result[x]}, "
+    f"y={fit_result[y]}, "
+    f"z={fit_result[z]}"
+)
+print(fit_result)
```
pckroon: Missing empty line

examples/mip/multiscenario.py
```diff
@@ -0,0 +1,42 @@
+from symfit import MIP, IndexedBase, Eq, Idx, Parameter, symbols, Sum
+import numpy as np
+
```
pckroon, line 11: For an example this needs a little paragraph describing the problem

examples/mip/multiscenario.py
```diff
@@ -0,0 +1,42 @@
...
+# Fixed costs for each plant
+data_fixed_costs = np.array([12000, 15000, 17000, 13000, 16000])
+
+# Transportation costs per thousand units
+data_trans_costs = np.array(
+    [[4000, 2000, 3000, 2500, 4500],
+     [2500, 2600, 3400, 3000, 4000],
+     [1200, 1800, 2600, 4100, 3000],
+     [2200, 2600, 3100, 3700, 3200]]
+)
```
pckroon, line 27: This is a min-cost-flow problem :)

examples/mip/multiscenario.py
```diff
@@ -0,0 +1,42 @@
...
+
+# Indices over the plants and warehouses
+plant = Idx('plant', range=len(data_capacity))
+warehouse = Idx('warehouse', range=len(data_demand))
+
+# Indexed variables. Initial values become coefficients in the objective function.
+open = IndexedBase(Parameter('open', binary=True, value=data_fixed_costs))
+transport = IndexedBase(Parameter('transport', value=data_trans_costs))
+capacity = IndexedBase(Parameter('capacity'))
+demand = IndexedBase(Parameter('demand'))
```
pckroon: `IndexedBase(Parameter(...))`, or `IndexedParameter(...)`?

examples/mip/multiscenario.py
```diff
@@ -0,0 +1,42 @@
...
+
+constraints = [
+    Sum(transport[warehouse, plant], warehouse) <= capacity[plant] * open[plant],
+    Eq(Sum(transport[warehouse, plant], plant), demand[warehouse])
+]
+
+data = {capacity[plant]: data_capacity, demand[warehouse]: data_demand}
+mip = MIP(constraints=constraints, data=data)
```
pckroon: This example does not really explain what's going on here I'm afraid.

examples/mip/multiscenario.py
```diff
@@ -0,0 +1,42 @@
...
+
+results = mip.execute()
+print(results)
+print(results[open])
+print(results[transport])
```
pckroon: Empty line

examples/mip/sudoku.py
```diff
@@ -0,0 +1,65 @@
...
+
+# Fix variables associated with cells whose values are pre-specified
+# lb = np.array([[0 if char == "." else 1 for j, char in enumerate(line)] for i, line in enumerate(grid)])
+lb = np.zeros((n, n, n), dtype=int)
+for i in range(n):
+    for j in range(n):
+        if grid[i][j] != '.':
+            v = int(grid[i][j]) - 1
+            lb[i, j, v] = 1
+ub = np.ones_like(lb)
```
pckroon, line 28: I don't like these variable names.
Some explanation as to why the one-hot encoding also wouldn't be misplaced in an example

symfit/api.py
```diff
@@ -15,5 +15,7 @@
 from symfit.core.argument import Variable, Parameter
 from symfit.core.support import variables, parameters, D
 
+from symfit.symmip import MIP
+
 # Expose the sympy API
 from sympy import *
```
pckroon, line 25: empty line

symfit/symmip/__init__.py
```diff
@@ -0,0 +1 @@
+from .mip import MIP
```
pckroon, line 1: empty line

symfit/symmip/backend/__init__.py
```diff
@@ -0,0 +1 @@
+from .gurobi import GurobiBackend
```
pckroon: empty line

symfit/symmip/backend/gurobi.py
```diff
@@ -0,0 +1,56 @@
+from dataclasses import dataclass, field
+
+import gurobipy as gp
+
+from sympy.printing.pycode import PythonCodePrinter
+from symfit.core.argument import Parameter
+
+
+VTYPES = {'binary': 'B', 'integer': 'I', 'real': 'C',}
```
pckroon: Should this be a GurobiBackend class variable?

symfit/symmip/backend/gurobi.py
```diff
@@ -0,0 +1,56 @@
...
+        return self.model.addConstr(constrexpr)
+
+    def set_objective(self, obj):
+        self.model.setObjective(obj)
+
+    @staticmethod
+    def param_vtype(p: Parameter):
+        """ Return the gurobi vtype for the parameter `p`. """
+        for key, vtype in VTYPES.items():
+            if p.assumptions0.get(key, False):
```
pckroon: `assumptions0`?

symfit/symmip/mip.py
```diff
@@ -0,0 +1,197 @@
...
+import numpy as np
+
+from symfit.symmip.backend import GurobiBackend
+from symfit.core.fit_results import FitResults
+from symfit.core.support import key2str
+
+
+DummyModel = namedtuple('DummyModel', 'params')
+
+VTYPES = {'binary': 'B', 'integer': 'I', 'real': 'C',} # ToDo: Enum?
```
pckroon: Seeing that this got duplicated here there has to be a better way

symfit/symmip/mip.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+class MIPBackend(Protocol):
```
pckroon, line 18: Should the GurobiBackend be a subclass of this?

symfit/symmip/mip.py
```diff
@@ -0,0 +1,197 @@
...
+    variables: List[Symbol] = field(init=False)
+    indexed_variables: List[Indexed] = field(init=False)
+
+    def __post_init__(self):
+        # Prepare constraints
+        self.constraints = [c if isinstance(c, Rel) else Eq(c, 0)
+                            for c in self.constraints]
+        free_symbols = reduce(operator.or_,
+                              (c.free_symbols for c in self.constraints),
+                              self.objective.free_symbols if self.objective else set())
```
pckroon: ... huh?

pckroon (changes requested):
Alright! I mostly like it. I do have some comments and questions though.

- Can you explain what the problem is with Models and Objectives + MIP? As I see it it should be possible to hook at least Model into this, maybe as MIPModel (see also ODEModel). I agree (for now) to leave out Objectives.
- What does this mean for Indexed<Things> in base symfit?
- I'm missing tests and docs ;)

pckroon:
Oh, and another question. I love the type annotations, but what does it mean for the minimum python version we can use/support?

tBuLi:
> Highs would be wonderful to add as it is open source.

Thanks for the tip, I wasn't aware of that package! I agree that we should have an open solver (as a default). But I tried to include it and found the documentation lacking and moreover, couldn't get the examples to work for Highs. So I've now instead added a wrapper for SCIP and made that the default, and that seems to work beautifully! So thanks for pointing me in this direction.

tBuLi:
> Alright! I mostly like it. I do have some comments and questions though.
> 
> * Can you explain what the problem is with Models and Objectives + MIP? As I see it it should be possible to hook at least Model into this, maybe as MIPModel (see also ODEModel). I agree (for now) to leave out Objectives.

Great question. The main issue is that in order to translate the symbolic definition into a MIP model, I actually need to interact directly with the symbolic definition and use a custom code printer to translate it into the target MIP backend. I found that with our existing models it is actually quite hard to get back to the symbolic expressions, because they are more designed to be easily called with data and parameters. Moreover, with the existing Model classes there is no control over the Printer being used; it is always the default one that sympy offers.

So my idea here was the following: within the `symmip` module I will make a new Model class, which has way less boilerplate code and allows direct control over the printing. Moreover, I will use the lessons from both `symfit` and the `slymfit` brainstorm session to make sure this model can also be dropped back into symfit eventually.

> * What does this mean for Indexed in base symfit?

A bit similar to the previous answer, I think that starting from scratch in a new submodule in a context that is full of indexed symbols will be a good opportunity to learn some best practices for dealing wit these objects, that can then be back ported into symfit.

> * I'm missing tests and docs ;)

Yeah it definitely still needs some work ;).

> Oh, and another question. I love the type annotations, but what does it mean for the minimum python version we can use/support?

I'm pretty sure that everything from >3.5 should be okay with this, so I don't expect any problems but let's keep our eyes open.

examples/mip/bilinear.py
tBuLi: Nope, it's horrible. The good news is that it is possible however, which means that we don't have to wrap everything even to get started. But we still need to think about a syntax to deal with this. Personally I think that it might be best if the backend class GurobiBackend accepts any options, which will then be passed on to gurobi without looking into it.

The example above would then become
```python
backend = GurobiBackend(NonConvex=2)
mip = MIP(- objective, constraints=constraints, backend=backend)

```

examples/mip/mip1.py
tBuLi: It was to show the syntax. Improved now, because I now run it with both backends to show how easy it is to switch.

examples/mip/multiscenario.py
tBuLi: That's the age old question isn't it ;). What I found is that it is better to stick to the sympy way, because if I made such an IndexedParameter, simply inheriting from IndexedBase wouldn't be enough because in sympy they will often explicitly return an instance of IndexedBase instead of `self.__class__`, thereby breaking subclassing. So it is not worth it imho to go that route.

symfit/symmip/backend/gurobi.py
tBuLi: It's what sympy gave us, so it'll have to do ;)

symfit/symmip/mip.py
tBuLi: agreed. I've now somewhat fixed this: in mip.py we only specify which types of variables there are as a list, whereas the specific backends can (should) dictate how these are to be translated to vtypes.

symfit/symmip/mip.py
tBuLi: For now I went with a protocol because it allows for static typing and duck-typing. An ABC was not worth it (yet) because there was no code reuse. But maybe I will make a common base class now that we have two solvers, but I'll keep the protocol nonetheless.

symfit/symmip/mip.py
tBuLi: Just taking the union of some sets of symbols! ;)

pckroon:
Alright, thanks for the explanation. If I understand it Model + Minimizer determine the Printer? Or even just Minimizer? Following this line of thought, does that mean *everything* should remain as symbolic as possible, right up until the Minimizer touches it? This also has implications for Objectives --- symbolical objectives have been on my wish list for a while. 

I think it's a good idea to fix this in code symfit Models now that we (you) know better, similar for Indexed\*. 

I think its fine to do some trial-and-error prototyping in a submodule like you do here, but I'd strongly prefer the final PR simply improving the symfit classes. This will also have the benefit of being able to use the tests to make sure you're not missing old use cases. I also think it'll result in less work since you'll be able to build on the existing framework. And where that doesn't fit, just break/adapt the old framework.



Can you do MIPODEModels? If no, your design is not good enough yet :D

examples/mip/bilinear.py
pckroon: That looks like an excellent option

examples/mip/multiscenario.py
pckroon: Alright. I'm not sure familiar with sympy, especially this corner. If this is how sympy does its thing I agree that we should do the same.

symfit/symmip/mip.py
pckroon: I need to read up on what `Protocol` does.

symfit/symmip/mip.py
pckroon: Oh right. I can see how this works now. At minimal this needs a comment. Also consider using `set.union` instead of `operator.or_`.

symfit/core/argument.py
```diff
@@ -93,6 +95,9 @@ class Parameter(Argument):
     _argument_name = 'par'
 
     def __new__(cls, name=None, value=1.0, min=None, max=None, fixed=False, **kwargs):
+        if 'binary' in kwargs:
+            kwargs['integer'] = kwargs['real'] = kwargs['nonnegative'] = True
```
pckroon, line 99:
```suggestion
            kwargs['integer'] = True
            kwargs['real'] = True
            kwargs['nonnegative'] = True
```

symfit/core/argument.py
```diff
@@ -115,13 +120,17 @@ def __init__(self, name=None, value=1.0, min=None, max=None, fixed=False, **assu
...
         self.fixed = fixed
 
-        if min is not None and max is not None and min > max:
-            if not self.fixed:
+        if min is not None and max is not None:
+            test = min > max
+            if isinstance(test, np.ndarray):
+                test = test.any()
+
+            if test and not self.fixed:
```
pckroon:
```suggestion
        if min is not None and max is not None:
            if np.any(min > max) and not self.fixed:
```

`np.any` does what you want if you pass it plain bools.

pckroon (changes requested):

Issue #374, opened 2024-10-25
Crash with numpy version 1.15 and above (function np.product removed)


kvr2007:
Hello!

First of all, great project, really wonderful to have it available as libre and open source code!

I was trying to run some of my scripts from 2022 in python3.12, and apparently numpy developers removed `np.product` at [some point](https://stackoverflow.com/a/49864286). So I get the following error when computing covariance matrix:

```
  File "...\site-packages\symfit\core\fit.py", line 574, in execute
    minimizer_ans.covariance_matrix = self.covariance_matrix(
                                      ^^^^^^^^^^^^^^^^^^^^^^^
  File ...\site-packages\symfit\core\fit.py", line 278, in covariance_matrix
    cov_matrix = self._covariance_matrix(best_fit_params,
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\site-packages\symfit\core\fit.py", line 256, in _covariance_matrix
    raw_dof = np.sum([np.product(shape) for shape in self.data_shapes[1]])
                      ^^^^^^^^^^
  File "...\site-packages\numpy\__init__.py", line 428, in __getattr__
    raise AttributeError("module {!r} has no attribute "
AttributeError: module 'numpy' has no attribute 'product'
```

If I change `np.product` to `np.prod` all works as expected. I think it's the only place in all of `symfit`, where `np.product` is used. So I would suggest to change the code. Alternatively, one needs to fix the numpy version.

artygo8:
I had this issue too, and in the meantime, as a workaround, you can do something like that:
```python
import numpy as np
from symfit import D, Eq, Fit, Model, parameters, variables

# MONKEY PATCH
np.product = np.prod
```

PR #375, opened 2024-10-29
Fix issue #374 (Crash with numpy version 1.15 and above (function np.product removed))

osvalB:
Fixes Issue #374. Based on the numpy version, fit.py decides if np.prod or np.product should be used.

### DIFF
--- a/symfit/core/fit.py
+++ b/symfit/core/fit.py
@@ -51,7 +51,7 @@ class TakesData(object):
         :param named_data: assign dependent, independent and sigma variables data by name.
 
         Standard deviation can be provided to any variable. They have to be prefixed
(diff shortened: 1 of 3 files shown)


tBuLi:
Thank you for supplying this fix! 

LooseVersion does raise the following warning: ```DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.``` Perhaps you could switch to that recommended solution?

And the automated tests also give ```ImportError: cannot import name 'simps' from 'scipy.integrate'```, it seems that simps has been renamed simpson... Would you mind applying the same patch for that import error?

osvalB:
Happy to help. I changed to 'packaging.version' and also changed 'simps' to 'simpson' according to the scipy version. Best, Os

